PR #56, opened 2023-09-11
MDTZ-941 Disassociate entity

jvrankul23:
This PR is currently inclusive of changes in https://github.com/atlassian-labs/figma-for-jira/pull/55 which should be merged first.

## Changes in this PR
* Added an `/entities/disassociateEntity` route and use case to handle disassociating Jira issues from Figma designs
* Standardised how ids are generated for `AtlassianDesign` entities. What's stored will always be a composite of `{fileKey}/{nodeId}`
* Consolidated `associateWith` and `disassociateFrom` types into a single `AtlassianEntity` type

### DIFF
--- a/src/web/routes/entities/index.ts
+++ b/src/web/routes/entities/index.ts
@@ -66,7 +66,7 @@ entitiesRouter.post(
 				atlassianUserId,
 				connectInstallation: res.locals.connectInstallation,
 			})
-			.then((design) => res.status(HttpStatusCode.Created).send({ design }))
+			.then((design) => res.status(HttpStatusCode.Ok).send({ design }))
 			.catch((error) => next(error));
 	},
 );
@@ -86,7 +86,7 @@ entitiesRouter.post(
 				atlassianUserId,
 				connectInstallation: res.locals.connectInstallation,
 			})
-			.then((design) => res.status(HttpStatusCode.Created).send({ design }))
+			.then((design) => res.status(HttpStatusCode.Ok).send({ design }))
 			.catch((error) => next(error));
 	},
 );
--- a/src/web/routes/entities/index.ts
+++ b/src/web/routes/entities/index.ts
@@ -5,8 +5,14 @@ import type {
 	AtlassianDesign,
 	ConnectInstallation,
 } from '../../../domain/entities';
-import type { AssociateEntityUseCaseParams } from '../../../usecases';
-import { associateEntityUseCase } from '../../../usecases';
+import type {
+	AssociateEntityUseCaseParams,
+	DisassociateEntityUseCaseParams,
+} from '../../../usecases';
+import {
+	associateEntityUseCase,
+	disassociateEntityUseCase,
+} from '../../../usecases';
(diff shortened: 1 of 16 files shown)


src/web/routes/entities/index.ts
```diff
@@ -23,6 +29,28 @@ type AssociateEntityResponse = Response<
...
+>;
+
+const getUserIdHeaderOrThrow = (
+	userId: string | string[] | undefined,
+	res: Response,
+): string => {
+	if (!userId || typeof userId !== 'string') {
+		const errorMessage = 'Missing or invalid User-Id header';
+		res.status(401).send(errorMessage);
+		throw errorMessage;
```
akostevich-atlassian, line 49: **issue:** Currently, `getUserIdHeaderOrThrow` does 2 things:

1. Sends an error response to a client
2. Throws an exception

I think we need to do either #1 or #2 but not both since it can cause tricky issues with request processing (e.g., when an error handler attempts to write to an already sent response). Also, it is better to avoid throwing a non-`Error` (e.g., `string`) to avoid type-related issues in error handling code. 

To address this, please, consider throwing a specific error (e.g., `BadRequestError`), which we could handle in `errorHandlerMiddleware`:

```ts
// web/errors.ts
export class BadRequestError extends Error {}


// src/web/routes/entities/index.ts
const getUserIdHeaderOrThrow = (
	userId: string | string[] | undefined,
	res: Response,
): string => {
	if (!userId || typeof userId !== 'string') {
		throw new BadRequestError('Missing or invalid User-Id header');
	}
	return userId;
}

// web/middleware/error-handler-middleware.ts
export const errorHandlerMiddleware = (
	err: Error,
	req: Request,
	res: Response,
	next: NextFunction,
): void => {
	// ...

	if (err instanceof BadRequesError) {
		res.status(HttpStatusCode.BadRequest).send(err.message);
	} 

	// ...
};
```

akostevich-atlassian (changes requested):
Amazing work, @jvrankul23, on this ! 🚀

src/web/routes/entities/index.ts
```diff
@@ -23,6 +29,28 @@ type AssociateEntityResponse = Response<
 	{ connectInstallation: ConnectInstallation }
 >;
 
+export type DisassociateEntityRequestParams = Omit<
```
atlasctz, line 32: Was curious what everyone's thoughts are on separating the exported types into a separate file?

src/web/routes/entities/index.ts
jvrankul23: Yeah great suggestion, this has simplified the code nicely as well.

src/web/routes/entities/index.ts
jvrankul23: I like it, keeps the route definitions tidy. Particularly for this route which had a few type definitions. I went ahead and made the change - if we want to apply this to all routes I can make the change in a separate PR.

atlasctz (approved):

akostevich-atlassian (approved):

PR #57, opened 2023-09-12
Mdtz 959 shadow delete from issue props

atlasctz:
**READY FOR REVIEW**

As the title says. 

There's one bigger change to the whole shadow read / write flow.

Apparently, axios already JSON.stringifies / parses requests / responses that have been specified as `Content-Type: application/json` in the headers, so there is no need to explicitly do this ourselves.

This is exactly the behaviour we want as Jira entity properties **must** be stored in valid JSON format. See: https://developer.atlassian.com/cloud/jira/platform/jira-entity-properties/#limitations-of-entity-properties

Quote: "The value stored in each property must be in valid JSON format. (Valid JSON format is defined as anything that [JSON.parse](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) can read)."

A small gotcha is nock automatically JSON.strigifies object bodies that are provided to its mock methods which imitates axios, but if you supply a string it assumes it is already stringified, so does not do it.
Therefore for `attached-design-url` (v1) issue property sets, which take a pure URL string, I had to manually JSON.stringify in the tests only.


src/infrastructure/jira/jira-client/jira-client.ts
```diff
@@ -109,15 +110,17 @@ class JiraClient {
...
 		try {
 			response = await axios.get<GetIssuePropertyResponse>(url.toString(), {
-				headers: new AxiosHeaders().setAuthorization(
-					this.buildAuthorizationHeader(url, 'GET', connectInstallation),
-				),
+				headers: new AxiosHeaders()
+					.setAuthorization(
+						this.buildAuthorizationHeader(url, 'GET', connectInstallation),
+					)
+					.setAccept('application/json'),
```
atlasctz, line 123: According to this [jira doc](https://developer.atlassian.com/cloud/jira/platform/jira-entity-properties/) entity properties must be of valid JSON format (anything that JSON.parse can read). 
[Get issue property](https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issue-properties/#api-rest-api-2-issue-issueidorkey-properties-propertykey-get) example also lists in the example to explicitly set the `Accept` header to `application/json`. So I have done this. Before, if this header wasn't set and a `string` was provided here, it would encode as text not json

src/infrastructure/jira/jira-service.ts
```diff
@@ -135,9 +134,8 @@ class JiraService {
 				connectInstallation,
 			);
 
-			const storedAttachedDesignUrlIssuePropertyValues = JSON.parse(
-				ensureString(response.value),
```
atlasctz, line 139: No longer necessary as axios already `JSON.parse` the response

src/infrastructure/jira/jira-service.ts
```diff
@@ -173,6 +171,100 @@ class JiraService {
...
+			if (response.value === design.url) {
+				await jiraClient.deleteIssueProperty(
+					issueIdOrKey,
+					propertyKeys.ATTACHED_DESIGN_URL,
+					connectInstallation,
+				);
+			}
+		} catch (error) {
+			if (error instanceof JiraClientNotFoundError) {
+				return; // Swallow not found errors
```
atlasctz, line 219: If a property to delete is not found, then we don't want to return that to the user. Perhaps it has already been deleted, in which case this is not an error. But maybe I should add a log here in case?

src/infrastructure/jira/jira-client/jira-client.ts
```diff
@@ -109,15 +110,17 @@ class JiraClient {
 		connectInstallation: ConnectInstallation,
 	): Promise<GetIssuePropertyResponse> => {
 		const url = new URL(
-			`/rest/api/2/issue/${issueIdOrKey}/properties/${propertyKey}`,
+			`${ISSUE_PROPERTY_BASE_URI}/${issueIdOrKey}/properties/${propertyKey}`,
```
akostevich-atlassian, line 113: **nitpick:** Technically, `ISSUE_PROPERTY_BASE_URI` is not a [URI](https://datatracker.ietf.org/doc/html/rfc3986) but a path. Therefore, consider renaming it to `ISSUE_PROPERTY_API_PATH`. This will also help to avoid confusion with `connectInstallation.baseUrl`.

Also, we could just inline the constant value into the string (`/rest/api/2/issue/${issueIdOrKey}/properties/${propertyKey}`) for the following reason:

- It gets easier to understand what endpoint it refers to since we don't need to jump among multiple places in the code.
- It is used just a few times within `JiraClient` and it neither a magic "string" nor subject to change. Therefore, it is perfectly fine to use it as is.

src/infrastructure/jira/jira-service.ts
```diff
@@ -135,9 +134,8 @@ class JiraService {
 				connectInstallation,
 			);
 
-			const storedAttachedDesignUrlIssuePropertyValues = JSON.parse(
-				ensureString(response.value),
-			) as AttachedDesignUrlV2IssuePropertyValue[];
+			const storedAttachedDesignUrlIssuePropertyValues =
```
akostevich-atlassian, line 137: **suggestion:** Taking into account that our Connect App does not have an exclusive ownership over these Issue Properties, it can contain any value. Therefore, it can be safer to validate the schema of their values.

Here is what [Jira docs](https://developer.atlassian.com/cloud/jira/platform/jira-entity-properties) say:

> Entity properties can be modified by all users that have permission to edit the entity. Entity properties can be modified by all apps in the system and exist in a global namespace.

> Be defensive when retrieving app properties, and don't assume data consistency (arbitrary keys may be modified or deleted by users).

src/infrastructure/jira/jira-service.ts
```diff
@@ -173,6 +171,100 @@ class JiraService {
...
+	 * @internal
+	 * Only visible for testing. Please use {@link deleteDesignUrlInIssueProperties}
+	 */
+	deleteAttachedDesignUrlInIssuePropertiesIfPresent = async (
+		issueIdOrKey: string,
+		design: AtlassianDesign,
+		connectInstallation: ConnectInstallation,
+	): Promise<void> => {
+		try {
+			const response = await jiraClient.getIssueProperty(
```
akostevich-atlassian, line 204: **nitpick:** It seems we could attempt to delete the Issue Property straightaway without fetching it first. Both `GET` and `DELETE` endpoints throw `JiraClientNotFoundError` if the Issue Property is not found.

akostevich-atlassian (changes requested):
Please, the comments above.

src/infrastructure/jira/jira-service.ts
atlasctz: No we can't do this since a user may link more than 1 design to the issue, and if you recall, the way the v1 property works (`attached-design-url`) is that the **first** linked design is stored. If further designs are linked, the v1 property does not update. So, only when the initially linked design is removed should we do the delete.

Example:
- A user creates an issue, and links design 1 as their first linked design
- `attached-design-url` _is set to design 1_
- User links design 2 to the issue
- `attached-design-url` _is not updated_
- User removes design 2 from the issue
- `attached-design-url` _is not deleted_
- User removes design 1 from the issue
- `attached-design-url` _is deleted_

src/infrastructure/jira/jira-service.ts
```diff
@@ -173,6 +171,100 @@ class JiraService {
...
+				issueIdOrKey,
+				propertyKeys.ATTACHED_DESIGN_URL_V2,
+				connectInstallation,
+			);
+
+			const storedAttachedDesignUrlIssuePropertyValues =
+				response.value as AttachedDesignUrlV2IssuePropertyValue[];
+
+			const issuePropertyValueToRemove: AttachedDesignUrlV2IssuePropertyValue =
```
jvrankul23, line 244: nit: Is this `const` necessary? Looks like we only read the `url` key in the below filter `issuePropertyValueToRemove.url`, we could probably just use `url` directly and optionally rename it in the args de-structuring, i.e.

```
{ url: urlToRemove, displayName }: AtlassianDesign,
```

```
const newAttachedDesignUrlIssuePropertyValue =
		storedAttachedDesignUrlIssuePropertyValues.filter(
			({ url }) => url !== urlToRemove,
		);
```

src/infrastructure/jira/jira-service.ts
akostevich-atlassian: Oh, I see! I've overlooked this. Thank you for the explanation, Costa!

akostevich-atlassian (approved):

src/infrastructure/jira/jira-service.ts
```diff
@@ -173,6 +171,100 @@ class JiraService {
...
+					url,
+					name: displayName,
+				};
+
+			const newAttachedDesignUrlIssuePropertyValue =
+				storedAttachedDesignUrlIssuePropertyValues.filter(
+					({ url }) => url !== issuePropertyValueToRemove.url,
+				);
+
+			await jiraClient.setIssueProperty(
```
jvrankul23, line 255: We should be able to safely no-op (and log) here if `issuePropertyValueToRemove.url` wasn't found in the issue property array

jvrankul23 (approved):

src/infrastructure/jira/jira-service.ts
atlasctz: Added validation

src/infrastructure/jira/jira-client/jira-client.ts
atlasctz: Removed `ISSUE_PROPERTY_BASE_URI` and just put the string in directly, I agree that there's no benefit to pulling it out.

src/infrastructure/jira/jira-service.ts
atlasctz: Done

src/infrastructure/jira/jira-service.ts
```diff
@@ -154,23 +187,128 @@ class JiraService {
...
+				propertyKeys.ATTACHED_DESIGN_URL_V2,
+				connectInstallation,
+			);
+		} catch (error) {
+			if (error instanceof JiraClientNotFoundError) {
+				return; // Swallow not found errors
+			}
+			throw error;
+		}
+		const storedAttachedDesignUrlIssuePropertyValues = JSON.parse(
```
akostevich-atlassian, line 267: **nitpick:** Currently, the code does _unsafe_ type assertion and then validation. It would be safer to do validation and then type assertion.

Interestingly, we could use [a type guard with an assertion signature](https://learntypescript.dev/07/l7-assertion-signature) to solve this problem (which was an interesting finding for me).

Please, consider the following:

```ts
deleteFromAttachedDesignUrlV2IssueProperties = async (
  issueIdOrKey: string,
  { url, displayName }: AtlassianDesign,
): Promise<void> => {
  // ...
  const storedAttachedDesignUrlIssuePropertyValues = JSON.parse(response.value);
  assertAttachedDesignUrlV2IssuePropertyValueType(storedAttachedDesignUrlIssuePropertyValues);

  // Type should be inferred by the TS compiler here thanks to the type guard with an assertion signature.
  // ...
}

private assertAttachedDesignUrlV2IssuePropertyValueType = (
  value: unknown,
) assertys value is AttachedDesignUrlV2IssuePropertyValue[] => {
  // ...
};
```

**PS:** Please, also pay attention that you do not `ensureString` since `JSON.parse` throws if any non-string input is provided.

atlasctz: Updated, thanks for the suggestion @akostevich-atlassian

akostevich-atlassian (approved):

PR #58, opened 2023-09-12
MDTZ-990/MDTZ-968 DB schema changes for webhooks flows

jv3nk:
This PR renames the `FigmaWebhook` table to `FigmaTeam` and adds a `clientKey` FK to the `ConnectInstallation` table, so we can look up Connect secrets when we receive an incoming webhook. It also adds an `AssociatedDesign` table, which will be used to look up associated design entities for a given fileKey/clientKey that will need to be updated when we receive a webhook event.

This should hopefully unblock most of the webhooks related tasks.


prisma/schema.prisma
```diff
@@ -8,12 +8,26 @@ datasource db {
...
 model ConnectInstallation {
-  id           Int    @id @default(autoincrement())
-  key          String
-  clientKey    String @unique
-  sharedSecret String
-  baseUrl      String
-  displayUrl   String
+  id               Int                @id @default(autoincrement())
+  key              String
+  clientKey        String             @unique
```
jv3nk, line 13: Now that this is being used as an FK, I'm wondering if it would be better to make this the primary key of this table.

prisma/schema.prisma
```diff
@@ -8,12 +8,26 @@ datasource db {
...
+  displayUrl       String
+  FigmaWebhook     FigmaTeam[]
+  AssociatedDesign AssociatedDesign[]
+}
+
+model AssociatedDesign {
+  id          Int     @id @default(autoincrement())
+  fileKey     String
+  nodeId      String?
+  isPrototype Boolean
```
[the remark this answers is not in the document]
jvrankul23: I believe files themselves also have a `nodeId` that is always `0:0`, so we should be able to safely store `nodeId` for all design entities

prisma/migrations/20230912072503_add_associated_designs_table/migration.sql
```diff
@@ -0,0 +1,16 @@
+-- CreateTable
+CREATE TABLE "AssociatedDesign" (
```
damoster, line 2: /nit probs too late to change this for existing tables already created (or is it?) but postgres table naming convention is snake case not camel case (reason being that its actually case insensitive so a lot of editors default lower case everything making it hard to read without snake case)

prisma/migrations/20230912072258_rename_figma_webhook_table_and_add_client_key_relation/migration.sql
```diff
@@ -0,0 +1,36 @@
...
+CREATE TYPE "FigmaTeamStatus" AS ENUM ('PENDING', 'ACTIVE', 'PAUSED', 'ERROR');
+
+-- DropTable
+DROP TABLE "FigmaWebhook";
+
+-- DropEnum
+DROP TYPE "FigmaWebhookStatus";
+
+-- CreateTable
+CREATE TABLE "FigmaTeam" (
```
damoster, line 17: /nit Same as https://github.com/atlassian-labs/figma-for-jira/pull/58/files#r1323743604

prisma/schema.prisma
damoster: Hold up, what are we actually using as the FK? I thought we would have only one table containing clientKey which other tables join on to get the value (meaning also that only one table needs to update that value and you can get those related tables to on delete cascade if desired). i.e.
```sql
CREATE TABLE connect_installation {
  id BIGSERIAL PRIMARY KEY,
  clientKey varchar,
  baseUrl varchar
  -- other columns
};

CREATE TABLE associated_design {
  connect_installation_id
    REFERENCES connect_installation ON DELETE CASCADE,
  fileKey varchar,
  nodeId varchar
  -- other columns
};

-- Select queries would look like
SELECT fileKey, nodeId
FROM associate_design
  JOIN connect_installation ON associate_design.connect_installation_id = connect_installation.id
WHERE connect_installation.clientKey = :clientKey;
```

prisma/schema.prisma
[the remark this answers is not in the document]
damoster: > I believe files themselves also have a nodeId that is always 0:0, so we should be able to safely store nodeId for all design entities

Can confirm, Jon and I verified this with Figma in meeting with them the other day

damoster (commented):
Overall is looking good except for the foreign key, should consider using an internal id as use for foreign key instead

prisma/schema.prisma
jv3nk: Good point, there's no need to use the `clientKey` as an FK here, can just use the `id` instead, I'll update.

akostevich-atlassian (approved):

prisma/schema.prisma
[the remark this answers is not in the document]
jv3nk: Thanks, I've made `nodeId` required.

jv3nk:
Updated the PR with the following changes:

- Identifiers are now [mapped](https://www.prisma.io/docs/concepts/components/prisma-schema/data-model#mapping-model-names-to-tables-or-collections) to `snake_case` with `@map`
- `nodeId` in the `AssociatedDesign` table is now required
- `ConnectInstallation` FKs now use the `id` column instead of `clientKey`
- Squashed migrations - after merging you'll need to run `npm run stop:sandbox && npm run start:sandbox` to re-initialise your local DB

prisma/schema.prisma
```diff
@@ -8,35 +8,62 @@ datasource db {
...
+model AssociatedDesign {
+  id          Int     @id @default(autoincrement())
+  fileKey     String  @map("file_key")
+  nodeId      String  @map("node_id")
+  isPrototype Boolean @map("is_prototype")
+
+  connectInstallation   ConnectInstallation @relation(fields: [connectInstallationId], references: [id], onDelete: Cascade)
+  connectInstallationId Int                 @map("connect_installation_id")
+
+  @@unique([fileKey, nodeId, isPrototype, connectInstallationId])
```
damoster, line 32: @jvrankul23 so are we going to say that two files are different if one is a prototype and one isn't? If that's not the intention, should we make this to instead be
```typescript
@@unique([fileKey, nodeId, connectInstallationId])
```

prisma/migrations/20230913005905_initial/migration.sql
```diff
@@ -0,0 +1,70 @@
...
+CREATE UNIQUE INDEX "figma_team_webhook_id_key" ON "figma_team"("webhook_id");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "figma_team_team_id_connect_installation_id_key" ON "figma_team"("team_id", "connect_installation_id");
+
+-- AddForeignKey
+ALTER TABLE "associated_design" ADD CONSTRAINT "associated_design_connect_installation_id_fkey" FOREIGN KEY ("connect_installation_id") REFERENCES "connect_installation"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "figma_team" ADD CONSTRAINT "figma_team_connect_installation_id_fkey" FOREIGN KEY ("connect_installation_id") REFERENCES "connect_installation"("id") ON DELETE CASCADE ON UPDATE CASCADE;
```
damoster, line 70: Why is prisma adding a default `ON UPDATE CASCADE`? 🤔 for foreign keys using internal identifiers, this would never happen, can we force prisma to only do `ON DELETE CASCADE` but not `ON UPDATE CASCADE`?

damoster (commented):
Looking good, just one more question to address regarding unique constraint

prisma/migrations/20230913005905_initial/migration.sql
jv3nk: Oh good catch, yeah Prisma defaults to `ON UPDATE CASCADE`, I've changed this to `ON UPDATE NO ACTION` (Postgres default behaviour).

prisma/schema.prisma
jvrankul23: @damoster From what I've been able to gather through testing, files cannot be prototypes (neither can nodes), a prototype is just a specific view in Figma files (similar to inspect mode vs. design mode). @stannnous and I were discussing this the other day and we may end up removing this `isPrototype` field and just storing prototypes as nodes, depending on how discussions with Figma go.

jv3nk: Thanks @jvrankul23, in that case I'll leave this as is while we're still storing `isPrototype`. Should hopefully be pretty easy to remove the column down the line if we need to.

jvrankul23 (approved):

PR #59, opened 2023-09-13
Add missing scopes and fix issue url

jvrankul23:
## Changes in this PR

* `file_dev_resources:write` is not inclusive of `file_dev_resources:read` - added this extra scope
* Jira `DELETE` scopes are required to delete issue properties - added this extra scope
* Correctly build `issueUrl` from issue API response and the `ConnectInstallation` baseUrl


src/infrastructure/figma/figma-service.ts
```diff
@@ -30,6 +30,10 @@ const extractDataFromFigmaUrlOrThrow = (url: string): FigmaUrlData => {
 	return urlData;
 };
 
+export const buildIssueTitle = (issueKey: string, issueSummary: string) => {
+	return `[${issueKey}] ${issueSummary}`;
```
jvrankul23, line 34: This is the shape of title that Figma's widget/plugin construct and store, applying it here as well for consistency.

src/infrastructure/figma/figma-service.ts
```diff
@@ -153,7 +159,10 @@ export class FigmaService {
 
 		if (response.errors?.length > 0) {
 			const errorMessage = response.errors.map((err) => err.error).join('|');
-			getLogger().error(errorMessage, 'Created dev resources with errors');
+			getLogger().error(
```
akostevich-atlassian, line 162: **question:** Do we need to throw an error after logging it? 

If no, consider logging just a message without an error creation:

```ts
getLogger().error('Created dev resources with errors', response.errors);
```

src/infrastructure/jira/jira-client/testing/mocks.ts
```diff
@@ -1,6 +1,7 @@
 import { v4 as uuidv4 } from 'uuid';
 
 import { Duration } from '../../../../common/duration';
+import { MOCK_ISSUE_KEY } from '../../../figma/testing';
```
akostevich-atlassian, line 4: **nitpick:** Our tests in `infrastructure/jira` depend on mocks from `infrastructure/figma/testing`. This can be an undesired dependency since these modules do not and should not not know about each other.

The issue happens because both `figma` and `jira` modules use the same `Issue` data. Therefore, in order to resolve this, consider moving domain-related constant  (e.g., `MOCK_ISSUE_KEY`, `MOCK_ISSUE_TITLE`) to `domain/testing` by analogue with `MOCK_ISSUE_ID`. As a result, both `infrastructure/jira` and `infrastructure/figma/testing` will either depend on their own mocks or mocks from `domain/testing`.

src/usecases/utils.ts
```diff
@@ -0,0 +1,3 @@
+export const buildIssueUrl = (baseUrl: string, issueKey: string) => {
```
akostevich-atlassian, line 1: **suggestion (non-blocking):** Consider moving this method to a lower layer: e.g., `infrastructure/jira/jira-utils/ts` can be a good fit.

akostevich-atlassian (approved):

src/infrastructure/figma/figma-service.ts
jv3nk: Agree that we don't need to create an error, but I think pino's logging API requires the object first unlike console.log ([docs](https://github.com/pinojs/pino/blob/master/docs/api.md#loggererrormergingobject-message-interpolationvalues)). You might also need to wrap the errors in an object so they get merged into the JSON log structure correctly

```typescript
getLogger().error({errors: response.errors}, 'Created dev resources with errors');
```

jv3nk (approved):

src/infrastructure/figma/figma-service.ts
atlasctz: I believe it needs to be inside an `err` property:
https://getpino.io/#/docs/api?id=mergingobject-object

so `{err: response.errors}`

jv3nk: I think those docs are just saying if you pass an `Error` as the first argument it'll get merged into the JSON log object under the `err` key. The first argument can be any object with arbitrary keys, and those will get merged into the final log object.

PR #61, opened 2023-09-14
Remove handling of prototype urls

jvrankul23:
Prototypes are simply a view mode in Figma, similar to dev mode and design mode. Therefore it does not make sense for us to store `type: PROTOTYPE` for designs fetched by the GET file or GET file nodes endpoints. Removed all code relating to prototypes.

### DIFF
--- a/src/infrastructure/figma/figma-service.test.ts
+++ b/src/infrastructure/figma/figma-service.test.ts
@@ -159,7 +159,6 @@ describe('FigmaService', () => {
 			const expectedEntity = transformNodeToAtlassianDesign({
 				fileKey: MOCK_FILE_KEY,
 				nodeId: MOCK_NODE_ID,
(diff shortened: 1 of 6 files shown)


jv3nk:
Noticed there's still a `PROTOTYPE` value in `AtlassianDesignType` that can be removed https://github.com/atlassian-labs/figma-for-jira/blob/MDTZ-1000-remove-prototype-handling/src/domain/entities/atlassian-design.ts#L14

jv3nk (approved):

jvrankul23:
@jv3nk I decided to leave that enum value in because technically we aren't removing it from our design provider schema yet. If Figma's implementation of prototypes changes, or they have something else that fits the bill of a prototype, they'll be aware of and can immediately start using this value

PR #64, opened 2023-09-15
MDTZ-970: Persist `AssociatedFigmaDesign` within the associate flow

akostevich-atlassian:
## Changes

- **[Improvement]:** Renames `AssociatedDesign` to  `AssociatedFigmaDesign` to reflect the fact that it contains Figma's identifiers. These Figma identifiers are required exclusively for the syncing functionality (for resolving associated `nodeId` by their `fileKey` on a webhook request).
- **[Feature]:** Introduces `FigmaDesignIdentity`, which represents a pair of Figma's `fileKey` and `nodeId`. It fully encapsulates the creation and parsing of Figma's identifiers. Currently, `FigmaDesignIdentity` provides:

  -  `FigmaDesignIdentity.fromFigmaDesignUrl` to create a new instance from the given design URL. Used in "/associateDesign" where the input is a design URL.  
  - `FigmaDesignIdentity.fromAtlassianDesignId` to create a new instance from the given Atlassian Design ID. Used in "/disassociateDesign" where the input is an Atlassian Design ID.
  - `FigmaDesignIdentity.prototype.toAtlassianDesignId` to convert itself to Atlassian Design ID. Used in data transformation and can be used for webhook processing.
 
  `FigmaDesignIdentity` enables upper-level services to expose a more concise API (e.g., a single `FigmaService.fetchByDesignId` instead of pair `FigmaService.fetchByDesignId` / `FigmaService.fetchByDesignUrl`).
- **[Feature]** Implements writing `AssociatedFigmaDesign` within "/associateDesign".
- **[Improvement]:** Do not include default`nodeId` (`0:0`) when build `AtlassianDesign.id` for a Figma file to avoid extra complexity and inconsistent behavior among flows.
- **[Fix]** Fix the issue with Figma's design URL parser not considering  presence of `:` and `%3A` (which is encoded `:`) in the `node-id` query parameter. Use `URL` API instead of regular expression for a more reliable parsing.


src/usecases/associate-entity-use-case.ts
```diff
@@ -21,9 +22,15 @@ export const associateEntityUseCase = {
 		atlassianUserId,
 		connectInstallation,
 	}: AssociateEntityUseCaseParams): Promise<AtlassianDesign> => {
+		const figmaDesignId = FigmaDesignIdentity.fromFigmaDesignUrl(entity.url);
+
 		const [design, issue] = await Promise.all([
-			figmaService.fetchDesignByUrl(entity.url, atlassianUserId),
+			figmaService.fetchDesignById(figmaDesignId, atlassianUserId),
 			jiraService.getIssue(associateWith.id, connectInstallation),
+			associatedFigmaDesignRepository.upsert({
```
jv3nk, line 30: What do you think about doing this upsert _after_ the design is sent to Data Depot? Otherwise if the ingestion fails, we will try to re-ingest this design if we receive a webhook for the containing file.

akostevich-atlassian: That's a very good point, Justin! We should definitely do this.

akostevich-atlassian: Fixed!

src/domain/entities/figma-design-identity.ts
```diff
@@ -0,0 +1,45 @@
+const FIGMA_DOCUMENT_NODE_ID = '0:0';
+
+export class FigmaDesignIdentity {
```
akostevich-atlassian, line 3: **note:** Unit tests are still missing for this class. Working on this now...

akostevich-atlassian: Unit tests have been added.

akostevich-atlassian:
@jv3nk , @jvrankul23 , the PR is ready for review. I apologise for the delay -- hopefully, it didn't block your work much.

src/domain/entities/figma-design-identity.test.ts
```diff
@@ -0,0 +1,139 @@
...
+		it('should throw when fileKey is empty string', () => {
+			expect(() => new FigmaDesignIdentity('')).toThrow();
+		});
+
+		it('should throw when nodeId is empty string', () => {
+			expect(() => new FigmaDesignIdentity(MOCK_FIGMA_FILE_KEY, '')).toThrow();
+		});
+	});
+
+	describe('fromFigmaDesignUrl', () => {
```
jv3nk, line 20: Should be `fromAtlassianDesignId`

src/domain/entities/figma-design-identity.test.ts
```diff
@@ -0,0 +1,139 @@
...
+		it('should return identity when Atlassian design ID does not contain node part', () => {
+			const result = FigmaDesignIdentity.fromAtlassianDesignId(
+				`${MOCK_FIGMA_FILE_KEY}/${MOCK_FIGMA_NODE_ID}`,
```
jv3nk, line 23: I think the Atlassian design ID isn't supposed to have a node ID for this test.

src/domain/entities/figma-design-identity.test.ts
```diff
@@ -0,0 +1,139 @@
...
+			);
+
+			expect(result).toStrictEqual(
+				new FigmaDesignIdentity(MOCK_FIGMA_FILE_KEY, MOCK_FIGMA_NODE_ID),
+			);
+		});
+
+		it('should throw when  Atlassian design ID has unexpected format', () => {
+			expect(() => FigmaDesignIdentity.fromFigmaDesignUrl(``)).toThrow();
```
jv3nk, line 42: Should be calling `fromAtlassianDesignId`

src/domain/entities/figma-design-identity.test.ts
```diff
@@ -0,0 +1,139 @@
...
+			).toString();
+
+			const result = FigmaDesignIdentity.fromFigmaDesignUrl(designUrl);
+
+			expect(result).toStrictEqual(
+				new FigmaDesignIdentity(MOCK_FIGMA_FILE_KEY, '42:1'),
+			);
+		});
+
+		it('should throw when URL has an unexpected format', () => {
```
jv3nk, line 95: /nit You could structure this as a parameterised test with Jest's [`it.each`](https://jestjs.io/docs/api#testeachtablename-fn-timeout)

```typescript
		it.each([
			`https://www.figma.com`,
			`https://www.figma.com/file`,
			`https://www.figma.com?param=file%2Fsome-id`,
		])('should throw when URL has an unexpected format', (url: string) => {
			expect(() => FigmaDesignIdentity.fromFigmaDesignUrl(url)).toThrow();
		});
```

jv3nk (changes requested):
Looks great, `FigmaDesignIdentity` has made things a lot cleaner. Just had some feedback on some of the tests, otherwise LGTM.

src/domain/entities/figma-design-identity.test.ts
akostevich-atlassian: Oh, great catch! I copied/pasted this test too many times and finally pasted an incorrect one 🙈

src/domain/entities/figma-design-identity.test.ts
akostevich-atlassian: Resolved!

src/domain/entities/figma-design-identity.test.ts
akostevich-atlassian: Resolved!

src/domain/entities/figma-design-identity.test.ts
akostevich-atlassian: Great suggestion! Updated the test accordingly.

jv3nk (approved):

src/domain/entities/figma-design-identity.ts
```diff
@@ -0,0 +1,59 @@
...
```
jvrankul23, line 3: nit: sorry @akostevich-atlassian I've only just had a chance to look at this. Thoughts on renaming to `FigmaDesignIdentitfier`?

akostevich-atlassian: No worries, Jon: your review is welcome anytime! :)

We could definitely explore other options! I called it `FigmaDesignIdentitfier` since it conceptually represents a _compound ID_ of a Figma design (`fileKey` + `nodeId`) -- it is a quite standard name for this kind of entities.

Do you have, Jon, any options in mind, so I could understand better in which direction you are thinking?

jvrankul23: Sorry might be getting terms confused, in code it's currently defined as `FigmaDesignIdentity`, I was thinking we could change it to `FigmaDesignIdentifier`. Which one do you prefer?

akostevich-atlassian: Ah, I see. Yes, I agree, `FigmaDesignIdentifier` will be a more precise option. Let me address this in a separate PR later today.

PR #65, opened 2023-09-15
MDTZ-971 Add endpoint for receiving webhook events

jv3nk:
This is now ready for review.

**Changes**

- I've updated the `FigmaTeamStatus` enum in the database to only have two values, `OK` and `ERROR`. Previously it had more values, some of which corresponded to the webhook status we receive from Figma, but this was confusing as it wasn't a 1:1 mapping. 
- Implemented the flow for receiving a webhook event. At a high level, it:
  - Checks if the event is a `FILE_UPDATE`, discards the event otherwise
  - Checks that we have valid OAuth2 credentials for the Figma team admin that created the webhook
    - If we don't have valid credentials, the `FigmaTeam` status is set to `ERROR`
  - Gets identifiers for all associated designs for the updated file from the database
  - Fetches designs from Figma using these identifiers
  - Submits the designs to Jira


src/infrastructure/jira/jira-client/jira-client.ts
```diff
@@ -62,6 +65,24 @@ class JiraClient {
 		return response.data;
 	};
 
+	deleteDesign = async (
+		designId: FigmaDesignIdentity,
+		connectInstallation: ConnectInstallation,
+	): Promise<FigmaDesignIdentity> => {
+		const url = new URL(
+			`/rest/designs/1.0/design/${designId.toAtlassianDesignId()}`,
```
jv3nk, line 73: @damoster is this the correct endpoint for deleting designs from DD?

damoster: Yes that's correct. Technically that's the public endpoint you've used here which connect apps call (i.e. you've done the right there but there's an internal equivalent for when internal services call DD such as the gdrive service)

src/infrastructure/jira/jira-service.ts
```diff
@@ -46,25 +50,32 @@ class JiraService {
...
+	updateDesigns = async (
+		designs: AtlassianDesign[],
+		connectInstallation: ConnectInstallation,
+	): Promise<void> => {
+		const response = await jiraClient.submitDesigns(
+			{
+				designs: designs.map((design) => ({
+					...design,
+					addAssociations: [],
+					removeAssociations: [],
```
damoster, line 65: /nit just fyi, we've designed the api in such a way that if there's no intention changing any associations. Can just leave both `addAssociations` and `removeAssociations` as null

jv3nk: Good to know! I'll switch the `[]`s to `null` and update our request body type to reflect this.

src/domain/entities/associated-design.ts
```diff
@@ -0,0 +1,22 @@
+export class AssociatedDesign {
```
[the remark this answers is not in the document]
jv3nk: Yup, forgot to delete this.

src/domain/entities/figma-team.ts
```diff
@@ -0,0 +1,17 @@
+export enum FigmaTeamStatus {
```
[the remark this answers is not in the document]
jv3nk: It's the status of a configured team, though it is a confusing at the moment since I've used some of the status values for Figma webhooks (`ACTIVE` and `PAUSED`). I'm thinking of reworking this to just include two statuses, `OK`, `ERROR`.

`OK` - the webhook for the team has been created and the stored OAuth2 credentials for the Figma team admin are valid
`ERROR` - The stored OAuth2 credentials for the Figma team admin are invalid, and the team will need to be re-configured

I don't think we can easily check if a webhook subscription is still active, so the `ERROR` status would just be used to represent invalid OAuth2 credentials.

src/usecases/webhook-use-case.ts
```diff
@@ -0,0 +1,152 @@
...
+
+/**
+ * Helper to wait for an array of Promises to be settled (fulfilled or
+ * rejected) and sort the results. Uses `Promise.allSettled` to await the
+ * Promises in parallel, then sorts the results into an object.
+ *
+ * @param promises The promises to await
+ * @returns A result object containing the fulfilled and rejected results
+ */
+const settlePromisesAndSortResults = async <T>(
```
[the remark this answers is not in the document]
jv3nk: Thanks for the suggestion, agree that this would be better served just handling the result grouping functionality and not calling `allSettled`.

src/infrastructure/figma/figma-service.ts
```diff
@@ -60,91 +47,59 @@ export class FigmaService {
 		}
 	};
 
-	fetchDesignByUrl = async (
-		url: string,
+	fetchDesignById = async (
```
jvrankul23, line 50: This looks like a great change, nice!

src/usecases/webhook-use-case.ts
```diff
@@ -0,0 +1,152 @@
...
+
+	await jiraService.updateDesigns(designs, connectInstallation);
+};
+
+const handleFileDeleteEvent = async (
+	associatedFigmaDesigns: AssociatedFigmaDesign[],
+	connectInstallation: ConnectInstallation,
+): Promise<void> => {
+	// Delete designs from Jira
+	const { rejected, fulfilled } = await settlePromisesAndSortResults(
```
jvrankul23, line 79: We may also need to think about the phone home (creating or deleting dev resources) and writing to issue properties as part of these webhook flows

src/domain/entities/figma-team.ts
jvrankul23, line 1: can we link out to Figma documentation for these statuses and/or add code comments to explain the enum values?

src/web/routes/webhooks/index.ts
```diff
@@ -0,0 +1,36 @@
...
+export const webhooksRouter = Router();
+
+webhooksRouter.post(
+	'/receive',
+	(req: TypedRequest<FigmaWebhookPayload>, res, next) => {
+		const { event_type, file_key, webhook_id } = req.body;
+
+		webhookUseCase
+			.execute(webhook_id, file_key, event_type)
+			// Figma docs specify that you should return a 200 OK if sucessful
```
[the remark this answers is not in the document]
jv3nk: My interpretation of the docs is that anything other than a `200` response would be considered unsuccessful.

src/usecases/webhook-use-case.ts
```diff
@@ -0,0 +1,152 @@
...
+		}
+	},
+};
+
+const handleFileUpdateEvent = async (
+	associatedFigmaDesigns: AssociatedFigmaDesign[],
+	figmaTeam: FigmaTeam,
+	connectInstallation: ConnectInstallation,
+): Promise<void> => {
+	const designs = await Promise.all(
```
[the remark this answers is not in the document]
jv3nk: Good question, if a lot of different nodes from a single file are associated with issues then this could fire off a large number of requests. I'm not sure how common this will be however, so I'm tempted to leave this as is for now and leave batching requests as an optimisation we can look at later if needed.

src/domain/entities/figma-team.ts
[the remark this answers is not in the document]
akostevich-atlassian: Got it! Thank you for the explanation, Justin!

I like the idea of restricting the enum to just simple 2 values. In addition to this, please, consider renaming the enum to `FigmaTeamAuthStatus` since it describes not the status of the whole team but its auth configuration.

```ts
export enum FigmaTeamAuthStatus {
  OK = 'OK',
  ERROR = 'ERROR',
}
```

If it is difficult to change or you have concerns, please, ignore :)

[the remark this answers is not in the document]
jv3nk: Good suggestion, I've updated the name and squashed the migrations.

prisma/schema.prisma
```diff
@@ -44,23 +44,21 @@ model FigmaOAuth2UserCredentials {
...
 
-enum FigmaTeamStatus {
-  PENDING
-  ACTIVE
-  PAUSED
+enum FigmaTeamAuthStatus {
+  OK
   ERROR
 
   @@map("figma_team_status")
```
akostevich-atlassian, line 51: **nitpick:** One more place to rename :)

src/infrastructure/jira/jira-service.ts
```diff
@@ -47,25 +51,32 @@ class JiraService {
...
+		this.throwIfSubmitDesignResponseHasErrors(response);
+	};
 
-		// TODO: Confirm whether we need to consider the use case below as a failure and throw or just leave a warning.
-		if (response.unknownIssueKeys?.length) {
-			throw JiraServiceSubmitDesignError.unknownIssueKeys(
-				response.unknownIssueKeys,
-			);
-		}
+	updateDesigns = async (
```
akostevich-atlassian, line 57: **suggestion:** Consider renaming the method to `submitDesigns` for the following reasons:

- Conceptually, this method does "upsert" but not the "update" operation.
- It will be consistent with the existing `submitDesign` (which does the same thing but for 1 design).

akostevich-atlassian: Also, consider refactor the existing `submitDesign` to delegate all the work to `submitDesigns` :)

src/infrastructure/repositories/figma-team-repository.ts
```diff
@@ -0,0 +1,74 @@
...
+		id: number,
+		status: FigmaTeamAuthStatus,
+	): Promise<void> => {
+		try {
+			await prismaClient.get().figmaTeam.update({
+				data: { status },
+				where: { id },
+			});
+		} catch (e: unknown) {
+			if (e instanceof PrismaClientKnownRequestError && e.code === 'P2025') {
```
akostevich-atlassian, line 49: **nitpick:** Justin, consider creating a constant with a self-descriptive name for this value. It would be also great to add a link to the documentation that describes this error code.

src/domain/entities/figma-team.ts
```diff
@@ -0,0 +1,16 @@
...
+	ERROR = 'ERROR',
+}
+
+export type FigmaTeam = {
+	readonly id: number;
+	readonly webhookId: string;
+	readonly teamId: string;
+	readonly teamName: string;
+	readonly figmaAdminAtlassianUserId: string;
+	readonly status: FigmaTeamAuthStatus;
```
akostevich-atlassian, line 12: **question:** Justin, do we need to rename this to `authStatus`?

src/domain/entities/testing/mocks.ts
```diff
@@ -137,3 +152,55 @@ export const generateJiraIssue = ({
 	self,
 	fields,
 });
+
+export const generateAssociatedFigmaDesignCreateParams = ({
+	designId = MOCK_FIGMA_DESIGN_IDENTITY,
```
akostevich-atlassian, line 157: **nitpick:** It seems it should be `generateFigmaDesignIdentity()`.

src/domain/entities/testing/mocks.ts
```diff
@@ -137,3 +152,55 @@ export const generateJiraIssue = ({
...
+	connectInstallationId = Math.floor(Math.random() * 10000),
+}: Partial<AssociatedFigmaDesignCreateParams> = {}): AssociatedFigmaDesignCreateParams => ({
+	designId,
+	connectInstallationId,
+});
+
+export const generateAssociatedFigmaDesign = ({
+	id = generateRandomInteger(),
+	designId = generateFigmaDesignIdentity(),
+	connectInstallationId = Math.floor(Math.random() * 10000),
```
akostevich-atlassian, line 167: **nitpick:** We could use a new awesome `generateRandomInteger`.

akostevich-atlassian (approved):

src/infrastructure/jira/jira-service.ts
jv3nk: Great suggestion, I've renamed the method to `submitDesigns` and it now takes an array of `SubmitDesignParams`. I've also updated `submitDesign` to just call out to the new `submitDesigns` method.

src/domain/entities/figma-team.ts
jv3nk: Agreed that it's a bit clearer, have renamed this column and updated the relevant methods/domain types.

src/domain/entities/figma-team.ts
jv3nk: This has been reworked to no longer be coupled to the Figma webhook statuses.

src/usecases/webhook-use-case.ts
```diff
@@ -0,0 +1,152 @@
...
+	figmaTeamRepository,
+	RepositoryRecordNotFoundError,
+} from '../infrastructure/repositories';
+
+const SUPPORTED_WEBHOOK_EVENT_TYPES: Set<WebhookEventType> = new Set([
+	WebhookEventType.FILE_UPDATE,
+	WebhookEventType.FILE_DELETE,
+]);
+
+export const webhookUseCase = {
```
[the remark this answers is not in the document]
jv3nk: Since we decided to not handle `FILE_DELETE` events, the use case has become a lot simpler so this shouldn't be necessary anymore.

PR #68, opened 2023-09-19
MDTZ-1013: Implement `checkAuth` action

akostevich-atlassian:
Implements the `checkAuth` action responsible for checking authorization status and returning an OAuth 2.0 authorization endpoint.

The change allows to _fully_ construct an OAuth 2.0 authorization endpoint on the Connect App side, and therefore apply some [OAuth 2.0 Security Best Practices](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics#name-insufficient-redirect-uri-v) (e.g., include an [anti-forgery state token](https://datatracker.ietf.org/doc/html/rfc6819#section-5.3.5) in an authorization endpoint URL for protection against Cross-Site Request Forgery attacks). The application of these best practices is out of scope of this PR.

`checkAuth` deprecates and replaces `check3LO` / `grant3LO` actions (which will be removed from the codebase as soon as Jira is updated accordingly).

## Changes

- **[Feature]** Implement the `checkAuth` action
- **[Improvement]** Construct Figma's authorization endpoint using [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) Web API to ensure that all parameters are correctly encoded.
- **[Improvement]** Rename `check3loUseCase` to `checkUserFigmaAuthUseCase` to add more context about its business purpose.
- **[Improvement]** Rename `addOAuthCredentialsUseCase` to `addFigmaOAuthCredentialsUseCase` to avoid confision between Atlassian and Figma auth.


stannnous (approved):

src/web/routes/auth/index.ts
```diff
@@ -48,26 +96,3 @@ authRouter.get(
...
 );
-
-/**
- * Checks whether the given Atlassian user is authorized to call Figma API.
- *
- * TODO: Replace with a link to public documentation.
- * @see https://hello.atlassian.net/wiki/spaces/MDT/pages/2796005496/RFC+-+Extending+generic+containers+specification+for+entity+associations#New-Provider-Actions
- */
-authRouter.get(
-	'/check3LO',
```
stannnous, line 59: Sorry I missed this earlier. We should keep the old route around so the app continues to work using the old action until all dependant PRs are released.

We can remove the old route as a follow up after the new action has been verified.

stannnous: You can use the new use case with the old route as the responses are compatible.

akostevich-atlassian: If I understand correctly, a Connect App descriptor serves as an additional level an abstraction, which allows to specify any endpoint for an action:

```
checkAuth: {
	templateUrl: `${
		getConfig().app.baseUrl
	}/auth/checkAuth?userId={userId}`,
},
grant3LO: {
	templateUrl: `...`,
},
check3LO: {
	templateUrl: `${
		getConfig().app.baseUrl
	}/auth/checkAuth?userId={userId}`, // `check3LO` points to `checkAuth` since  its response is backwards compatible.
},
```

Therefore, as long as we keep the `check3LO` action in the app descriptor and the action points out to the backwards compatible endpoint, we are on the safe side.

Please, let me know if I am missing something.

stannnous: You're right. That seems like a good approach except I realised old installations will not have the new descriptor so to avoid breaking them when the app backend deploys to staging, it must still serve the old routes.

akostevich-atlassian: Oh, you are right, Sam! Thank you for the explanation and a great catch.

✅  `check3LO` has been returned.

stannnous (approved):

jv3nk (approved):

PR #71, opened 2023-09-20
NONE Remove global unused mock check to fix test flakiness

jv3nk:
This was causing flakiness in our integration tests that expect an error from a Promise.all call. If multiple endpoints inside a Promise.all call are mocked, there's no guarantee that all of the mocks will be hit if one of the Promises rejects, since Promise.all immediately halts execution on a single rejection.

### DIFF
--- a/scripts/setup-jest.ts
+++ b/scripts/setup-jest.ts
@@ -31,18 +31,10 @@ afterEach(() => {
 });
 
 /**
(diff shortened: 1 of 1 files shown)


akostevich-atlassian (approved):

akostevich-atlassian:
**praise:** Great finding, Justin!

PR #73, opened 2023-09-21
MDTZ-977 Add schema validation for incoming requests

jv3nk:
This PR aims to refactor how we type our routes, so we can have better type safety/validation in our handlers.

- Route request/response types and schemas are now located in the route package, in `types.ts` and `schemas.ts` respectively
- Added schemas for all endpoint inputs (POST bodies and query parameters), which are used to validate incomings requests with `assertSchema`
- Added a middleware to extract the user's Atlassian user ID from request headers
- Removed `TypedRequest` in favour of using type parameters on the Express `Request` type instead
  - This gets us better type safety because it lets us type query parameters and locals as well


src/infrastructure/figma/schemas.ts
jv3nk, line 1: This file was split into the following files:

- `FigmaWebhookEventPayload` moved to `src/web/routes/figma/types.ts`
- `FigmaWebhookEventType` moved to `src/usecases/types.ts`
  - This was not moved to `src/web/routes/figma/types.ts` because the type is used in our use cases and I wanted to avoid use cases depending on the routes package
- `FIGMA_WEBHOOK_EVENT_TYPE_SCHEMA` moved to `src/usecases/schemas.ts`
- `FIGMA_WEBHOOK_PAYLOAD_SCHEMA` moved to `src/web/routes/figma/schemas.ts`

src/web/middleware/extract-user-id-from-headers-middleware.ts
```diff
@@ -0,0 +1,20 @@
+import type { NextFunction, Request, Response } from 'express';
+
+import { UnauthorizedError } from './errors';
+
+import { ensureString } from '../../common/stringUtils';
+
+export const extractUserIdFromHeadersMiddleware = (
```
jv3nk, line 7: I originally had this middleware also attempt to extract the user ID from the `userId` query parameter, but decided it was better to just do this in the `/checkAuth` route instead (which is the only place where we expect a `userId` in query params).

src/web/routes/entities/index.ts
```diff
@@ -2,47 +2,48 @@ import { HttpStatusCode } from 'axios';
...
 	'/associateEntity',
-	authHeaderSymmetricJwtMiddleware,
 	(
-		req: TypedRequest<AssociateEntityRequestParams>,
+		req: AssociateEntityRequest,
 		res: AssociateEntityResponse,
 		next: NextFunction,
 	) => {
-		const atlassianUserId = getUserIdHeaderOrThrow(req.headers['user-id']);
+		assertSchema(req.body, ASSOCIATE_ENTITY_REQUEST_BODY_SCHEMA);
```
akostevich-atlassian, line 40: **praise:** Amazing! I feel much more confident in what our service does with this assertion.

src/infrastructure/figma/schemas.ts
akostevich-atlassian: **suggestion:** Great reorganization, Justin!

Frankly speaking, I think we could move `FigmaWebhookEventType` to `src/web/routes/figma/types.ts` as well, but, as you mentioned, we need to remove the dependency on this type from `handleFigmaFileChangeEventUseCase` first.

There are 2 easy ways how we could do this:

1. Currently, `FigmaWebhookEventType` is used to determine whether we are interested in a received event.  We could move this check to the `/figma/webhook` route, so our use case do not need to bother about irrelevant events at all.  
  ```ts
  figmaRouter.post('/webhook', (req, res, next) => {
	  assertSchema(req.body, FIGMA_WEBHOOK_PAYLOAD_SCHEMA);
  
	  const { event_type, file_key, webhook_id } = req.body;
  
          if (event_type !== 'FILE_UPDATE') {
	          return res.sendStatus(HttpStatusCode.Ok)
          }
  
	  handleFigmaFileChangeEventUseCase
		  .execute(webhook_id, file_key, event_type)
		  // Figma docs specify that you should return a 200 OK if sucessful
		  .then(() => res.sendStatus(HttpStatusCode.Ok))
		  .catch(next);
  });
  ```
  Another benefit of this approach is that we could delegate different events to different use cases (which can have a very different business logic and implementation!) in a route.
  ```ts
  figmaRouter.post('/webhook', (req, res, next) => {
	  assertSchema(req.body, FIGMA_WEBHOOK_PAYLOAD_SCHEMA);
  
	  const { event_type, file_key, webhook_id } = req.body;
  
	  switch (event_type) {
		  case 'FILE_UPDATE': {
			  return handleFigmaFileUpdateEventUseCase
				  .execute(webhook_id, file_key, event_type)
				  .then(() => res.sendStatus(HttpStatusCode.Ok))
				  .catch(next);
		  }
		  case 'FILE_DELETE': {
			  // ...
		  }
		  default: return res.sendStatus(HttpStatusCode.Ok)
	  }
  });
  ```
	
2. Introduce a separate `SupportedFigmaFileChange` type for `handleFigmaFileChangeEventUseCase` with values our use case is interested in (currently, only `FILE_UPDATE`).

Off the top of my head, between these 2 options, I would go with option #1.

akostevich-atlassian:
**praise:** There is so much great stuff in this PR. I cannot wait to see it in `main` :)

akostevich-atlassian (approved):

src/web/middleware/extract-user-id-from-headers-middleware.ts
```diff
@@ -0,0 +1,20 @@
...
+	req: Request,
+	res: Response<unknown, { atlassianUserId: string }>,
+	next: NextFunction,
+): void => {
```
jvrankul23, line 11: :chefskiss:

jvrankul23:
Awesome PR, thanks @jv3nk. Feels a lot cleaner and safer now.

jvrankul23 (approved):

src/infrastructure/figma/schemas.ts
jv3nk: Great suggestion, I agree that option 1 looks better, I'll make this change.

PR #75, opened 2023-09-21
Add existing figma web panel

jvrankul23:
Adds static assets from the existing figma-for-jira Connect app and adds the webPanels module, controllable by a feature flag.


Roystbeef (approved):
LGTM!

PR #76, opened 2023-09-22
MDTZ-1017: Remove `AssociatedFigmaDesign` record on `/disassociateEntity` request

akostevich-atlassian:
## Changes

- **[Feature]:** Add `AssociatedFigmaDesign.associatedWithAri` field to track which Atlassian entity (e.g., Issue) a design is associated with. This knowledge is used to find and delete only a target `AssociatedFigmaDesign` record on a `/disassociateEntity` request.
- **[Feature]:** Update `/disassociateEntity` flow to delete `AssociatedFigmaDesign`.
- **[Improvement]:** Do not set up nock and Prisma for unit tests but only for integration tests.
- **[Improvement]:** Clean the database before each integration test.
- **[Improvement]:** Remove excessive integration tests for `/associateEntity` and `/disassociateEntity`.
- **[Improvement]:** Add unit test for `/associateEntity` and `/disassociateEntity`.

### DIFF
--- a/scripts/setup-jest-integration-tests.ts
+++ b/scripts/setup-jest-integration-tests.ts
@@ -13,6 +13,13 @@ beforeAll(() => {
 	});
 });
 
+beforeEach(async () => {
+	await prismaClient.get().associatedFigmaDesign.deleteMany({});
+	await prismaClient.get().figmaTeam.deleteMany({});
+	await prismaClient.get().figmaOAuth2UserCredentials.deleteMany({});
+	await prismaClient.get().connectInstallation.deleteMany({});
+});
+
 /**
  *  After each test, check for requests that went through
  *  without being mocked by nock.
@@ -28,13 +35,6 @@ afterEach(() => {
 		unmatchedRequests = [];
 		throw new Error(`${numberOfUnmatchedRequests} Unmatched Requests`);
 	}
-});
-
-beforeEach(async () => {
-	await prismaClient.get().associatedFigmaDesign.deleteMany({});
-	await prismaClient.get().figmaTeam.deleteMany({});
-	await prismaClient.get().figmaOAuth2UserCredentials.deleteMany({});
-	await prismaClient.get().connectInstallation.deleteMany({});
 });
 
 /**
--- a/scripts/setup-jest-integration-tests.ts
+++ b/scripts/setup-jest-integration-tests.ts
@@ -30,11 +30,11 @@ afterEach(() => {
 	}
 });
 
-beforeEach(() => {
-	prismaClient.get().associatedFigmaDesign.deleteMany({});
-	prismaClient.get().figmaTeam.deleteMany({});
-	prismaClient.get().figmaOAuth2UserCredentials.deleteMany({});
-	prismaClient.get().connectInstallation.deleteMany({});
+beforeEach(async () => {
+	await prismaClient.get().associatedFigmaDesign.deleteMany({});
+	await prismaClient.get().figmaTeam.deleteMany({});
+	await prismaClient.get().figmaOAuth2UserCredentials.deleteMany({});
+	await prismaClient.get().connectInstallation.deleteMany({});
 });
 
 /**
--- a/src/infrastructure/figma/figma-service.ts
+++ b/src/infrastructure/figma/figma-service.ts
@@ -2,10 +2,7 @@ import { AxiosError, HttpStatusCode } from 'axios';
 
 import { FigmaServiceCredentialsError } from './errors';
 import { figmaAuthService } from './figma-auth-service';
-import type {
-	CreateDevResourcesRequest,
-	CreateDevResourcesResponse,
-} from './figma-client';
+import type { CreateDevResourcesRequest } from './figma-client';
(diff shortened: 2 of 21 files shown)


src/infrastructure/figma/figma-service.ts
```diff
@@ -96,7 +93,7 @@ export class FigmaService {
 		issueKey: string;
 		issueTitle: string;
 		atlassianUserId: string;
-	}): Promise<CreateDevResourcesResponse> => {
+	}): Promise<void> => {
```
akostevich-atlassian, line 96: **note:** We don't want to expose Figma API model from the service. The return type can be safely replaces with `void` since it is not used.

src/infrastructure/repositories/associated-figma-design-repository.ts
```diff
@@ -22,53 +21,75 @@ export class AssociatedFigmaDesignRepository {
...
+		const records = await prismaClient.get().associatedFigmaDesign.findMany({
 			where: { fileKey, connectInstallationId },
 		});
 
-		return result.map(this.mapToDomainModel);
+		return records.map(this.mapToDomainModel);
 	};
 
-	deleteByDesignIdAndConnectInstallationId = async (
+	findByDesignIdAndAssociatedWithAriAndConnectInstallationId = async (
```
akostevich-atlassian, line 50: **note:** Required for integration tests only.

scripts/setup-jest-integration-tests.ts
```diff
@@ -30,6 +30,13 @@ afterEach(() => {
 	}
 });
 
+beforeEach(async () => {
```
jv3nk, line 33: Is this necessary? When we run integration tests with `npm run test:it`, we run [`prisma migrate reset`](https://www.prisma.io/docs/concepts/components/prisma-migrate/migrate-development-production#reset-the-development-database), which should delete any data from tests.

If we want to do this in code so that it's done on every integration test run (not just when we use `npm run test:it`), we could execute the reset command as a child process. Or maybe look at the workarounds suggested in https://github.com/prisma/prisma/issues/5230

src/infrastructure/repositories/associated-figma-design-repository.ts
```diff
@@ -22,53 +21,75 @@ export class AssociatedFigmaDesignRepository {
...
 			});
 
-			return this.mapToDomainModel(result);
+			return this.mapToDomainModel(record);
 		} catch (e: unknown) {
 			if (
 				e instanceof PrismaClientKnownRequestError &&
 				e.code === PrismaErrorCode.RecordNotFound
 			) {
-				throw new RepositoryRecordNotFoundError(e.message);
```
jv3nk, line 71: /nit Not sure about this `deleteXXX` method behaving differently to others, but I get we don't really have a naming convention for delete methods that don't throw on missing records. The return type does kind of indicate this, so this is probably okay.

jv3nk (approved):
Nice one, another flow down! Just had one question about resetting the DB, otherwise looks good.

scripts/setup-jest-integration-tests.ts
jv3nk: Just realised this is executing after each test so it's not doing the same thing as `npm run test:it`. I'm happy to merge this as is.

src/infrastructure/repositories/associated-figma-design-repository.ts
akostevich-atlassian: That's a good question, Justin!

What do you think about using the following convention?

- `deleteXxx` returns a deleted record or `null` if it does not exist. Likely, this is what we need in the majority of cases due to the idempotency requirement and concurrency -- usually, we want to make sure that a record is deleted but not necessarily delete it with a particular method call. If we look at other team's projects, all our `deleteXxx` methods provide this contract and use this convention.
- `deleleXxxOrThrowIfMissing returns a deleted record or throw if a record is missing. This is less likely but still possible contract for our methods.

Please, let me know, Justin, whether you have any concerns or find these names confusing or clunky.

scripts/setup-jest-integration-tests.ts
akostevich-atlassian: Justin, if I understand correctly, our `npm run test:it` currently takes care of cleaning data and applying the migrations (which is the primary task!) before running tests.

However, it can be also beneficial to clean data before _each_ test to prevent stale data (e.g., from a crashed test) to impact other tests. We already do this for some integration tests: please, see [here](https://github.com/atlassian-labs/figma-for-jira/blob/add6f10ae30a58dc14987fa9c3026abb52beb988/src/web/routes/entities/integration.test.ts#L272) for an example.

The PR just makes this automatic and more reliable. Please, pay attention that we do the same thing in other projects.

src/infrastructure/repositories/associated-figma-design-repository.ts
akostevich-atlassian: Added integration tests for `AssociatedFigmaDesignRepository.deleteByDesignIdAndAssociatedWithAriAndConnectInstallationId` to cover its tricky error handling.

PR #77, opened 2023-09-22
MDTZ-973 Add endpoint for configuring a webhook for a team

jv3nk:
~~This is mostly done, there's a few unit tests missing but I'm putting this up for review now since I won't be in on Monday.~~
Now ready for review

- Added the `/team/configure` endpoint
  - This accepts a `teamId` and creates a new webhook and `FigmaTeam` record
  - Originally this was supposed to accept a list of `teamId`s, but the most recent designs only let users configure one team at a time. We can adapt this to accept multiple `teamId`s if needed in a separate PR.
- When we create the webhook, we can add a `passcode` value that will be sent with every incoming webhook event. This can be used to authenticate webhook event requests
  - ~~For now I'm setting this to the SHA256 hash of the team admin's Atlassian user ID + their Figma team ID + the Connect installation secret~~ We now generate a UUID and save this against the `FigmaTeam` record
  - I've also added validation of the passcode to the `/figma/webhook` endpoint


src/infrastructure/jira/jira-client/jwt-utils.ts
```diff
@@ -9,7 +9,7 @@ export type JwtTokenParams = {
 		readonly query?: Record<string, unknown>;
 		readonly body?: Record<string, unknown>;
 	};
-	readonly connectClientKey: string;
+	readonly connectAppKey: string;
```
jv3nk, line 12: From the Connect [docs](https://developer.atlassian.com/cloud/bitbucket/understanding-jwt-for-apps/#claims) (emphasis mine):

> The issuer of the claim. Connect uses it to identify the application making the call. For example:
> **If the Atlassian product is the calling application: contains the unique identifier of the tenant. This is the clientKey** that you receive in the installed callback. You should reject unrecognized issuers.
> **If the app is the calling application: the app key specified in the app descriptor**.

We were correctly using the app key to generate JWTs, but the variable was mislabelled in this method.

src/infrastructure/figma/figma-service.ts
```diff
@@ -159,6 +164,49 @@ export class FigmaService {
...
+			endpoint: `${getConfig().app.baseUrl}/figma/webhook`,
+			passcode,
+			description: 'Figma for Jira Cloud',
+		};
+
+		const result = await figmaClient.createWebhook(request, accessToken);
+		return { webhookId: result.id, teamId: result.team_id };
+	};
+
+	validateWebhookPasscode = (passcode, input: WebhookPasscodeInput) => {
```
jv3nk, line 194: Need to add tests for the passcode generation/validation.

src/usecases/configure-figma-teams-use-case.ts
```diff
@@ -0,0 +1,29 @@
...
+		const { webhookId, teamId: figmaTeamId } =
+			await figmaService.createFileUpdateWebhook(
+				teamId,
+				atlassianUserId,
+				connectInstallation.sharedSecret,
+			);
+		await figmaTeamRepository.upsert({
+			webhookId,
+			teamId: figmaTeamId,
+			teamName: 'TODO',
```
jv3nk, line 23: We're still deciding how to handle team names, there's an ongoing LDR. This column may actually not be required, so I'm just leaving this as 'TODO' for now.

src/infrastructure/figma/figma-service.ts
```diff
@@ -159,6 +164,49 @@ export class FigmaService {
...
+	/**
+	 * @internal
+	 * Visible for testing only.
+	 */
+	generateWebhookPasscode = ({
+		atlassianUserId,
+		figmaTeamId,
+		connectInstallationSecret,
+	}: WebhookPasscodeInput): string => {
+		const input = atlassianUserId + figmaTeamId + connectInstallationSecret;
```
akostevich-atlassian, line 207: **suggestion (non-blocking):** It might be better to keep a Connect App installation shared secret exclusively between the Connect App and Jira.

Here is what we could do instead:

1. Generate a UUID and store it in a new `FigmaTeam.webhookPasscode` column. This solution also requires querying a database on each webhook request.
2. Sign `FigmaTeam.teamId` (which is passed to `POST/v2/webhooks`) with a public key stored in the environment variables. This solution does not require querying a database on each webhook request.

Let's discuss this when you are available, Justin. Anyway, it is not a blocker for this PR -- we can address this in a separate ticket (if needed).

src/usecases/handle-figma-file-update-event-use-case.ts
```diff
@@ -8,8 +8,29 @@ import {
...
 
 export const handleFigmaFileUpdateEventUseCase = {
-	execute: async (webhookId: string, fileKey: string): Promise<void> => {
+	execute: async (
+		webhookId: string,
+		fileKey: string,
+		passcode: string,
+	): Promise<void> => {
 		const figmaTeam = await figmaTeamRepository.getByWebhookId(webhookId);
+		const [connectInstallation, associatedFigmaDesigns] = await Promise.all([
```
akostevich-atlassian, line 17: **suggestion:** Consider query required data only after authorizing a request to reduce a risk of a denial-of-service attack. 

In order to authorize a request, we have to call `figmaTeamRepository.getByWebhookId(webhookId)` but a potentially expensive `associatedFigmaDesignRepository.findManyByFileKeyAndConnectInstallationId` call is not required for authorization.

While a sequential execution can slightly increase the latency, it can be a safer solution and should not be an issue for a background task (which is, normally, not time-sensitive).

src/usecases/configure-figma-teams-use-case.test.ts
```diff
@@ -0,0 +1,77 @@
...
+		await expect(
+			configureFigmaTeamUseCase.execute(
+				uuidv4(),
+				uuidv4(),
+				generateConnectInstallation(),
+			),
+		).rejects.toStrictEqual(error);
+	});
+
+	it('should throw if creating the FigmaTeam record fails', async () => {
```
akostevich-atlassian, line 60: **suggestion:** Verifying that out code does not catch an exception can be excessive since this is the default behavior, which does not require any special considerations. Consider testing only those error scenario where we _handle_ an error (e.g., try to recover from a failure, translate an error, suppress an error, etc.). Otherwise, we would need to add a test for almost any call of a built-in or our function (since many of them can throw an error).

For `configureFigmaTeamUseCase`, we might also need to test that a `FigmaTeam` record is _not_ created if a webhook creation fails (which is important from the business perspective).

src/web/routes/entities/integration.test.ts
```diff
@@ -52,16 +52,25 @@ import {
...
 	key: 'com.figma.jira-addon-dev',
 	clientKey: '4561b8be-e38b-43d4-84d9-f09e8195d117',
 	sharedSecret: '903b6b9e-b82b-48ea-a9b2-40b9e700df32',
 });
-const ASSOCIATE_JWT_TOKEN =
-	'JWT eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE2OTM5NTcyMDUsImV4cCI6NjAwMDAwMDE2OTM5NTcxNDQsImlzcyI6IjQ1NjFiOGJlLWUzOGItNDNkNC04NGQ5LWYwOWU4MTk1ZDExNyIsInFzaCI6IjQ2ZDE3MDU4OWU0MjM2Y2U0YTQ5MTFlMGQ1YWE4YjdkOWYzZjNlODZlN2E0ZTgzMzFhM2MyNWE5NTI0MWNjMmYifQ.E71S-uGRlVmEY8-iEEj4bl3SiOcDUlJ-36XGoq8tDHE';
-const DISASSOCIATE_JWT_TOKEN =
-	'JWT eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE2OTQ0MzU0NjcsImV4cCI6NjAwMDAwMDE2OTQ0MzU0MDAsImlzcyI6IjQ1NjFiOGJlLWUzOGItNDNkNC04NGQ5LWYwOWU4MTk1ZDExNyIsInFzaCI6ImEzYTcwOGIxNDQwMjdlM2U2ZWRjYWYzY2MzZTBkNzYzYWRhMzI4NDgwMTFjNzMzNGQwMjRkZGE2ZGQ5OWU2NWUifQ.j_Bb4M76LpldZglEY7wQE7KY1KsMIsuvTyqppQd2wBY';
+
+const ASSOCIATE_JWT_TOKEN = `JWT ${generateInboundRequestJwtToken({
```
akostevich-atlassian, line 63: **praise:** Amazing!

src/web/routes/configure/configure-router.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export const configureRouter = Router();
+
+configureRouter.use(
+	extractUserIdFromHeadersMiddleware,
+	authHeaderSymmetricJwtMiddleware,
+);
+
+configureRouter.post(
+	'/team',
```
akostevich-atlassian, line 26: **suggesgtion:** The containing folder and the router are called `configure` but this name can be a too generic and does not provide the context about what we configure here. If we need to add to the system something else configurable down the road, this folder might be not a good place for the new code (since it would contain configurable but _unrelated_ things).

Therefore, consider:

1. Renaming the containing folder from `configure` to `team` to add more context.
2. Renaming the endpoint from `configure/team` to `team/configure`.

src/web/routes/figma/integration.test.ts
```diff
@@ -219,6 +225,19 @@ describe('/figma', () => {
 				);
 			});
 
+			it('should return a 200 if the passcode is invalid', async () => {
```
akostevich-atlassian, line 228: **suggestion:** Per Figma's security recommendation, we need to return `HTTP 400` in this case:

> We would recommend comparing the passcode we pass back to you with the passcode originally provided when creating the endpoint to make sure they match before acting on the webhook trigger. If you receive a request with the wrong passcode, you should respond with a 400 Bad Request HTTP response which will immediately stop the webhook. ([source](https://www.figma.com/developers/api#webhooks-v2-security)).

akostevich-atlassian (approved):

akostevich-atlassian:
Another piece of the system is almost ready! 🚀  

Really great job on this, Justin!

src/infrastructure/figma/figma-client/figma-client.ts
```diff
@@ -192,6 +194,28 @@ export class FigmaClient {
...
+	 * Creates a new Figma webhook using the POST webhook endpoint
+	 *
+	 * @see https://www.figma.com/developers/api#webhooks-v2-post-endpoint
+	 */
+	createWebhook = async (
+		request: CreateWebhookRequest,
+		accessToken: string,
+	): Promise<CreateWebhookResponse> => {
+		const response = await axios.post<CreateWebhookResponse>(
+			`${getConfig().figma.apiBaseUrl}/v2/webhooks`,
```
akostevich-atlassian, line 208: **nitpick:** While testing other webhook-related changes, I noticed that we need the `webhooks:write` scope to call this API.

We need to add this scope to [config.figma.scope](https://github.com/atlassian-labs/figma-for-jira/blob/2b04d0e6dd4d2d6c08790f663a91b2def7555e6a/src/config/config.ts#L48).

src/infrastructure/figma/types.ts
```diff
@@ -0,0 +1,5 @@
+export type WebhookPasscodeInput = {
```
akostevich-atlassian, line 1: **suggestion:** Consider moving this type to `figma-service.ts` since it represents an utility type to describe the parameters of `FigmaService` method. We could even make this type private to that module.

src/web/routes/figma/integration.test.ts
jv3nk: Interesting, I wonder if `stop the webhook` means it no longer fires events. If it does, I think I would prefer to return a `200` to avoid potential for abuse. Will do some testing on this.

src/infrastructure/figma/figma-client/figma-client.ts
jv3nk: Great catch, I totally missed this. I'll make sure to e2e test this before it gets merged.

src/web/routes/figma/integration.test.ts
akostevich-atlassian: Hey Justin! This was my assumption too but I was not able to reproduce this behavior -- I keep receiving events after `HTTP 400`. We could double-check this with the Figma team.

src/infrastructure/figma/types.ts
jv3nk: Now that we're generating a UUID for the passcode, this is no longer needed.

src/infrastructure/figma/figma-service.ts
jv3nk: We've decided to go with option 1 here, to generate a UUID for the passcode and store it in the `FigmaTeam` table.

src/usecases/handle-figma-file-update-event-use-case.ts
```diff
@@ -8,8 +9,14 @@ import {
...
+	execute: async (figmaTeam: FigmaTeam, fileKey: string): Promise<void> => {
+		const [connectInstallation, associatedFigmaDesigns] = await Promise.all([
+			connectInstallationRepository.get(figmaTeam.connectInstallationId),
+			associatedFigmaDesignRepository.findManyByFileKeyAndConnectInstallationId(
+				fileKey,
+				figmaTeam.connectInstallationId,
+			),
+		]);
 
 		// Ensure team admin OAuth2 credentials are still valid
```
jv3nk, line 21: Just realised this block probably belongs to the `validateWebhookPayloadUseCase` now, will move it.

akostevich-atlassian (approved):

PR #78, opened 2023-09-25
MDTZ-993: Clean up Figma webhooks in the "Uninstall" lifecycle event

akostevich-atlassian:
## Changes

- **[Feature]:** Updates the `uninstalledUseCase` to remove Figma webhooks and `FigmaTeam` records.
- **[Improvement]:** Add test utilities for lifecycle event integration tests. 
- **[Improvement]:** Add integration tests for the `/lifecycleEvents/intstalled` and `/lifecycleEvents/unintstalled` endpoint.


src/usecases/uninstalled-use-case.ts
```diff
@@ -1,7 +1,28 @@
...
+		await Promise.all(
+			figmaTeams.map((figmaTeam) =>
+				figmaService.deleteWebhook(
+					figmaTeam.webhookId,
+					figmaTeam.figmaAdminAtlassianUserId,
+				),
+			),
+		);
+
 		await connectInstallationRepository.deleteByClientKey(clientKey);
```
jv3nk, line 26: What do you think about always deleting the `ConnectInstallation` (which will cascade delete `FigmaTeam` and `AssociatedDesign` records), even if deleting webhooks fails? Since uninstall events aren't retried, I don't think we'll get another opportunity to delete the `ConnectInstallation`, at least while don't have some queue implemented.

If we always delete the `ConnectInstallation`, we will also fail any received webhook events that we couldn't unsubscribe from. So hopefully Figma would delete the webhook when it never receives a successful response for the webhook event request.

src/web/routes/lifecycle-events/integration.test.ts
```diff
@@ -1,12 +1,243 @@
...
+		.delete(`/v2/webhooks/${webhookId}`)
+		.reply(status);
+};
 
 describe('/lifecycleEvents', () => {
-	it('/installed should respond 401 when JWT token is missing', () => {
-		return request(app)
-			.post('/lifecycleEvents/installed')
-			.expect(HttpStatusCode.Unauthorized);
+	describe('/installed', () => {
```
jv3nk, line 56: Thanks for adding these, think we've got integration tests on all our routes now 🎉

src/web/testing/mocks.ts
```diff
@@ -1,23 +1,84 @@
...
-import type {
-	ConnectInstallation,
-	ConnectInstallationCreateParams,
-} from '../../domain/entities';
+import { generateKeyPair } from 'node:crypto';
+import { promisify } from 'util';
 
 /**
- * Generates a JWT token that can be used to authorise inbound requests in
+ * Generates an asymmetric JWT token that can be used to authorise lifecycle event requests in
```
jv3nk, line 13: /nit Might be worth mentioning that you also need to mock the public key endpoint, and/or linking to https://developer.atlassian.com/cloud/jira/platform/security-for-connect-apps/#validating-installation-lifecycle-requests

jv3nk (approved):
👏

src/web/testing/mocks.ts
akostevich-atlassian: Great suggestion!

✅ Done!

src/usecases/uninstalled-use-case.ts
akostevich-atlassian: Excellent point, Justin! I was trying to design the use case for idempotency (in case the Figma team would like to add a queue) but I totally agree that we need to consider what we have right now (no queue and no retries).

I've updated the implementation and added a comment with more information about possible issues.

PR #79, opened 2023-09-26
MDTZ-1029: Use `BigInt` for primary keys in Prisma schema

akostevich-atlassian:
## Changes

- **[Feature]:** Uses [BigInt](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#bigint) for primary keys in Prisma schema (translated to Postrgres `bigserial`). Update the domain model to use `string` for the corresponding fields to avoid limitations imposed by BigInt (see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#use_within_json) for more detail).
- **[Improvement]:** Add `dbModel` testing utility for generating a `BigInt` ID. Update tests and mocks accordingly.
- **[Improvement]:** Apply minor refactoring to repositories for consistency. Prevent an ability to provide an ID to the `upsert` methods -- always generate IDs on the database side.

### DIFF
--- a/prisma/migrations/20230922041702_initial/migration.sql
+++ b/prisma/migrations/20230922041702_initial/migration.sql
@@ -3,7 +3,7 @@ CREATE TYPE "figma_team_auth_status" AS ENUM ('OK', 'ERROR');
 
 -- CreateTable
 CREATE TABLE "connect_installation" (
-    "id" SERIAL NOT NULL,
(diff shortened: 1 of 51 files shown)


jv3nk (approved):
Nice work! Hopefully this is the last schema change we need 🤞

PR #80, opened 2023-09-27
NONE Update webhook event payload schema and fix `jwt:generate` script

jv3nk:
Couple of small bugfixes:

- The first is to update the webhook event payload schema to make `file_key` and `file_name` optional for `PING` events
  - When we subscribe to a webhook, some `PING` events are emitted immediately. These events don't have a `file_key` or `file_name`, so our schema validation was failing and we were returning a `400`
- The second fix is just to update the `create-jwt.ts` script to fix a renamed parameter

### DIFF
--- a/src/infrastructure/webhook/webhook-service.test.ts
+++ b/src/infrastructure/webhook/webhook-service.test.ts
@@ -74,6 +74,7 @@ describe('WebhookService', () => {
 		jest
 			.spyOn(figmaTeamRepository, 'getByWebhookId')
 			.mockResolvedValue(figmaTeam);
+		jest.spyOn(figmaTeamRepository, 'updateAuthStatus').mockResolvedValue();
 		jest
 			.spyOn(figmaService, 'getValidCredentialsOrThrow')
 			.mockRejectedValue(new Error('auth error'));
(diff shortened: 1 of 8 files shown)


src/infrastructure/webhook/webhook-service.test.ts
```diff
@@ -74,6 +74,7 @@ describe('WebhookService', () => {
 		jest
 			.spyOn(figmaTeamRepository, 'getByWebhookId')
 			.mockResolvedValue(figmaTeam);
+		jest.spyOn(figmaTeamRepository, 'updateAuthStatus').mockResolvedValue();
```
jv3nk, line 77: Realised I forgot to mock this call, not sure how the test was passing before 🤔

akostevich-atlassian: Haha, nice catch! You, likely, had your test database running, so this unit test wrote to the database.

src/web/routes/figma/schemas.ts
```diff
@@ -43,12 +58,19 @@ export const FIGMA_WEBHOOK_PAYLOAD_SCHEMA: JSONSchemaTypeWithId<FigmaWebhookEven
...
 			'event_type',
-			'file_key',
-			'file_name',
 			'passcode',
 			'protocol_version',
 			'retries',
 			'timestamp',
 			'webhook_id',
 		],
+		if: {
```
akostevich-atlassian, line 67: Advance Ajv techniques are detected :)

akostevich-atlassian (approved):
**praise:** 👏 👏 👏

PR #81, opened 2023-09-27
MDTZ-1033 Set team name on FigmaTeam record

jv3nk:
This PR adds the business logic for setting the `teamName` on a `FigmaTeam`. When we configure a team, we fetch the team name from the [get team projects endpoint](https://www.figma.com/developers/api#get-team-projects-endpoint), and set this in the newly created `FigmaTeam` record.

To keep this team name in sync with Figma, we also fetch and update the team name whenever we handle a `FILE_UPDATE` event for said team. This is done on a best effort basis, and if we're unable to sync the name for whatever reason, we still proceed with the re-ingestion of designs.


src/infrastructure/figma/figma-client/types.ts
```diff
@@ -108,3 +108,13 @@ export type CreateWebhookResponse = {
 	readonly description: string | null;
 	readonly protocol_version: string;
 };
+
+export type Project = {
+	id: string;
```
akostevich-atlassian, line 113: **suggestion:** Consider adding the `readonly` modifier for consistency with other models.

src/usecases/handle-figma-file-update-event-use-case.test.ts
```diff
@@ -35,8 +38,56 @@ describe('handleFigmaFileUpdateEventUseCase', () => {
 			}),
 		);
 
+		it('should log a warning and continue if saving the team name to the database fails', async () => {
+			const associatedAtlassianDesigns: Record<string, AtlassianDesign> =
```
akostevich-atlassian, line 42: **nitpick:** Consider using `Map` instead of an `object` since it provides a handy API for its construction, which can simplify the code:

```ts
const associatedAtlassianDesignsById = new Map(
				associatedFigmaDesigns
					.map(figmaDesign => figmaDesign.designId.toAtlassianDesignId())
					.map(id => [id, generateAtlassianDesign({ id })])
				);
```

akostevich-atlassian (approved):

src/usecases/handle-figma-file-update-event-use-case.test.ts
jv3nk: Nice, I didn't know about `Map`

PR #82, opened 2023-09-27
NONE: Address project structure and naming inconsistencies

akostevich-atlassian:
## Changes

- **[Improvement]** Renames the `auth/callback` endpoint to `/figma/oauth/callback` to add more context about its responsibility. Moves the endpoint implementation from `authRouter` to `figmaRouter` to  bring Figma callbacks together (since they share similar properties: e.g., they came from Figma and they do not require Jira auth).
- **[Improvement]** Renames `WebhookService` to `FigmaWebhookService` (and related types) since it is specific to Figma and cannot be reused in other context. Moves `FigmaWebhookService`  to `infrastructure/figma`. @jv3nk , sorry for confusing you with my idea to put the service directly in `infrastructure` -- your initial solution was very reasonable. 
- **[Improvement]** Renames `figmaService.createDevResource` to  `figmaService.createDevResourceForJiraIssue` to provide more context about its responsibility.
- **[Improvement]** Moves `buildIssueUrl` from `infrastructure/jira/jira-utils.ts` to `domain` to avoid dependencies from use cases on utility functions from the `infrastructure` package. This function is not "exempted" from the dependency rule since it does not represent a service (for which we allow such dependencies for practical reasons). @jvrankul23 , sorry for confusing you with my request to move this function to `infrastructure` some time ago -- I was wrong 🙈 
- **[Improvement]** Applies minor refactoring to the contact of several methods.


src/domain/entities/jira-issue.ts
```diff
@@ -16,3 +16,7 @@ export type JiraIssue = {
 		readonly summary: string;
 	};
 };
+
+export const buildJiraIssueUrl = (baseUrl: string, issueKey: string) => {
```
akostevich-atlassian, line 20: **note:** We could make this function a static method of `JiraIssue` but this would require converting `JiraIssue` to a class. I've left this small utility function here for now for simplicity.

jv3nk (approved):
👍 ✨

PR #83, opened 2023-09-27
MDTZ-1032: Bind `FigmaOAuth2UserCredentials` to `ConnectInstallation` to enable its cascading deletion on "Uninstalled" event

akostevich-atlassian:
## Changes

- **[Feature]:** `FigmaOAuth2UserCredentials.connectInstallationId` to bind `FigmaOAuth2UserCredentials` to the corresponding `ConnectInstallation`. 
- **[Feature]:** Enables cascading deletion of `FigmaOAuth2UserCredentials` on the corresponding `ConnectInstallation` deletion (which happens on an "Uninstalled" lifecycle event).
- **[Feature]:** Introduces `ConnectUserInfo` to the domain as a handy value object holding a pair of Atlassian User ID and Connect Installation ID. It is used by several services and use cases.
- **[Improvement]:** Moves functions for mocking endpoints (using Nock) to `web/testing/xxx-api-mocks.ts` to make them reusable among tests. There is some work needs to be done to remove inconsistencies among these functions -- it will be addressed in a separate PR.


src/infrastructure/figma/figma-auth-service.ts
```diff
@@ -70,13 +76,30 @@ export class FigmaAuthService {
 			client_id: getConfig().figma.clientId,
 			redirect_uri: redirectUri,
 			scope: getConfig().figma.scope,
-			state: atlassianUserId, // TODO: MDTZ-1014: Use an anti-forgery state token to prevent Cross-Site Request Forgery (CSRF) attacks.
+			state: `${user.connectInstallationId}/${user.atlassianUserId}`, // TODO: MDTZ-1014: Use an anti-forgery state token to prevent Cross-Site Request Forgery (CSRF) attacks.
```
akostevich-atlassian, line 79: **note:** This is a naive approach of bypassign `connectInstallationId` / `atlassianUserId` through the OAuth 2.0 flow. Hopefully, we can improve this part within MDTZ-1014 by encrypting (and, probably, signing the pair) with some private key.

akostevich-atlassian:
@jv3nk , please, could I ask you to have a look at this PR when you have time? I would like to confirm the overall direction before starting working on tests :)

As you mentioned, "hopefully, it will be the last schema change" :D

jv3nk:
@akostevich-atlassian this approach LGTM, happy to go ahead 🙂

akostevich-atlassian:
@jv3nk , the PR is ready for review. Sorry for the delay: if you wish to merge your changes first -- please, let me know and just do it :)

jv3nk:
> @jv3nk , the PR is ready for review. Sorry for the delay: if you wish to merge your changes first -- please, let me know and just do it :)

Awesome, I've only got https://github.com/atlassian-labs/figma-for-jira/pull/86 open. I'd prefer to get this merged in first, the merge conflicts should be pretty easy to resolve on my branch. Looks like there's some conflicts with `main` to look at first though 😅

src/infrastructure/webhook/webhook-service.ts
```diff
@@ -29,9 +29,10 @@ export class WebhookService {
 
 		// Ensure team admin OAuth2 credentials are still valid
 		try {
-			await figmaService.getValidCredentialsOrThrow(
-				figmaTeam.figmaAdminAtlassianUserId,
-			);
+			await figmaService.getValidCredentialsOrThrow({
```
jv3nk: /nit We construct a `ConnectUserInfo` from `FigmaTeam` in a few places, would it be worth having a helper to do this? Could be a method on `FigmaTeam` or `ConnectUserInfo`? Though they would need to be converted to classes, so maybe it's not worth the effort.

src/usecases/check-user-figma-auth-use-case.ts
```diff
@@ -2,11 +2,18 @@ import {
 	figmaService,
 	FigmaServiceCredentialsError,
 } from '../infrastructure/figma';
+import { ConnectInstallation } from '../domain/entities';
 
 export const checkUserFigmaAuthUseCase = {
-	execute: async (atlassianUserId: string): Promise<boolean> => {
+	execute: async (
+		atlassianUserId: string,
```
jv3nk, line 9: /nit I think we could just accept a `ConnectUserInfo` param here?

src/web/testing/connect-api-mock.ts
jv3nk, line 1: Awesome job collating our API mocks, we had a lot of repetition/inconsistency, this should make writing tests a lot easier.

akostevich-atlassian: This actually took the majority of time I spent on this PR :D

jv3nk (approved):

src/infrastructure/webhook/webhook-service.ts
akostevich-atlassian: That's a good point -- I agree, Justin! Let me address this in a separate PR.

src/usecases/check-user-figma-auth-use-case.ts
akostevich-atlassian: Technically, we can but I'm afraid that it can inconsistent with other use cases. Many use cases accept the entire `ConnectIntallation` (since it is automatically fetched by middleware for use). This use case contract allows to avoid an extra query to the database for use cases that require a `ConnectIntallation` (e.g., to call Jira API).

Therefore, it can be more consistent to leave it as is and consistently accept a `ConnectIntallation` in all our use cases (but it is, definitely, not a hard rule).

Please, let me know if you have concerns about this, Justin.

PR #86, opened 2023-09-29
MDTZ-1019 / MDTZ-1034 Refactor handle file update event use case

jv3nk:
This PR resolves two tickets for the webhook events endpoint:

**MDTZ-1019** 
Updated how we fetch design data to only make one API request, passing in a list of node IDs. Previously we were firing an API request for each `AssociatedFigmaDesign` record.

To facilitate this, the `FigmaService.fetchDesignsByIds` method was added that accepts a `FigmaDesignIdentifier[]` and parses the file response into an `AtlassianDesign[]`.

**MDTZ-1034**
Updated error handling to update FigmaTeam auth status whenever we get a FigmaServiceCredentialsError when hitting Figma APIs.

### DIFF
--- a/src/infrastructure/figma/figma-service.test.ts
+++ b/src/infrastructure/figma/figma-service.test.ts
@@ -2,7 +2,7 @@ import type { AxiosResponse } from 'axios';
 import { AxiosError, HttpStatusCode } from 'axios';
 import { v4 as uuidv4 } from 'uuid';
 
-import { FigmaServiceCredentialsError } from './errors';
+import { FigmaServiceCredentialsError, FigmaServiceError } from './errors';
 import {
 	figmaAuthService,
 	NoFigmaCredentialsError,
@@ -17,8 +17,10 @@ import type {
 import { figmaClient, FigmaClientNotFoundError } from './figma-client';
 import {
 	generateChildNode,
+	generateFrameNode,
 	generateGetFileResponse,
 	generateGetFileResponseWithNode,
+	generateGetFileResponseWithNodes,
 } from './figma-client/testing';
 import {
 	buildDevResourceNameFromJiraIssue,
@@ -205,9 +207,6 @@ describe('FigmaService', () => {
 			jest
 				.spyOn(figmaService, 'getValidCredentialsOrThrow')
 				.mockResolvedValue(credentials);
-			jest
-				.spyOn(figmaAuthService, 'getCredentials')
-				.mockResolvedValue(credentials);
 			jest.spyOn(figmaClient, 'getFile').mockRejectedValue(mockError);
 
 			await expect(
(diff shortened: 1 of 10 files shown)


src/web/routes/figma/integration.test.ts
```diff
@@ -140,7 +152,7 @@ describe('/figma', () => {
 		describe('FILE_UPDATE event', () => {
 			let connectInstallation: ConnectInstallation;
 			let figmaTeam: FigmaTeam;
-			let associatedFigmaDesigns: AssociatedFigmaDesign[];
```
jv3nk, line 143: We were using this previously in tests to generate `AtlassianDesign`s / `FileResponse`s. This array was populated as we upserted `AssociatedFigmaDesign` records in the `beforeEach` block:

```typescript
associatedFigmaDesigns.push(
	await associatedFigmaDesignRepository.upsert(
		associatedFigmaDesignCreateParams,
	),
```

This was causing tests to fail when I switched over to using `FigmaService.fetchDesignsByIds`, because the order of `associatedFigmaDesigns` now needed to match the order of associated designs returned by `AssociatedFigmaDesignRepository.findManyByFileKeyAndConnectInstallationId`, so that `nock` can correctly match query params / request body.

To fix this, I now query `associatedFigmaDesigns` from the database in each test instead.

src/infrastructure/figma/figma-service.test.ts
```diff
@@ -202,24 +204,124 @@ describe('FigmaService', () => {
...
+			const result = await figmaService.fetchDesignsByIds(
+				[],
+				ATLASSIAN_USER_ID,
+			);
+
+			expect(result).toStrictEqual([]);
+			expect(figmaService.getValidCredentialsOrThrow).not.toBeCalled();
+		});
+
+		it('should throw a FigmaServiceError if design ids have different file keys', async () => {
```
akostevich-atlassian, line 291: **praise:** Great test cases!

src/infrastructure/figma/figma-service.ts
```diff
@@ -86,6 +86,53 @@ export class FigmaService {
...
+		designIds: FigmaDesignIdentifier[],
+		atlassianUserId: string,
+	): Promise<AtlassianDesign[]> => {
+		if (!designIds.length) {
+			return [];
+		}
+
+		// Ensure all design identifiers have the same file key
+		const fileKey = designIds[0].fileKey;
+		for (const designId of designIds.slice(1)) {
```
akostevich-atlassian, line 99: **nitpick:** Consider using `Array.prototype.every`, which could help to simplify the code:

```ts
const fileKey = designIds[0].fileKey;
const haveSameFileKey = designIds.every((designId) => designId.fileKey === fileKey);

if (!haveSameFileKey) {
	throw new FigmaServiceError('designIds must all have the same fileKey');
}
```

src/web/routes/figma/integration.test.ts
```diff
@@ -200,31 +209,43 @@ describe('/figma', () => {
...
 
 			it('should fetch and submit the associated designs to Jira', async () => {
-				const entries = associatedFigmaDesigns.map(
-					generateDesignAndFileResponseAndAtlassianDesign,
+				const associatedFigmaDesigns =
+					await associatedFigmaDesignRepository.findManyByFileKeyAndConnectInstallationId(
+						fileKey,
+						connectInstallation.id,
+					);
+				const nodeIds = associatedFigmaDesigns.flatMap(({ designId }) =>
```
akostevich-atlassian, line 217: **suggestion:** Consider using `map`/`filter` as a potentially simpler solution:

```ts
const nodeIds = associatedFigmaDesigns
  .map(({ designId }) => designId.nodeId)
  .filter(Boolean);
```

akostevich-atlassian (approved):
**praise:** Fantastic changes, Justin!

src/infrastructure/figma/figma-service.ts
jv3nk: TIL `Array.prototype.every`

PR #87, opened 2023-09-29
NONE: Add `FigmaTeam.adminInfo` to simplify `FigmaTeam` consumer code

akostevich-atlassian:
The PR addresses https://github.com/atlassian-labs/figma-for-jira/pull/83#discussion_r1340880224 from another PR.

## Changes

- **[Improvement]:** Adds a `FigmaTeam.adminInfo` property.
- **[Improvement]:** Update the `FigmaTeam` consumer code to leverage a new property instead of a manual construction of `ConnectUserInfo` object.

### DIFF
--- a/src/domain/entities/figma-team.ts
+++ b/src/domain/entities/figma-team.ts
@@ -7,17 +7,34 @@ export enum FigmaTeamAuthStatus {
 
 export class FigmaTeam {
 	private readonly _adminInfo: ConnectUserInfo;
+	readonly id: string;
+	readonly webhookId: string;
+	readonly webhookPasscode: string;
+	readonly teamId: string;
+	readonly teamName: string;
+	readonly figmaAdminAtlassianUserId: string;
+	readonly authStatus: FigmaTeamAuthStatus;
+	readonly connectInstallationId: string;
 
-	constructor(
-		readonly id: string,
-		readonly webhookId: string,
-		readonly webhookPasscode: string,
-		readonly teamId: string,
-		readonly teamName: string,
-		readonly figmaAdminAtlassianUserId: string,
-		readonly authStatus: FigmaTeamAuthStatus,
-		readonly connectInstallationId: string,
-	) {
+	constructor(params: {
+		id: string;
+		webhookId: string;
+		webhookPasscode: string;
+		teamId: string;
+		teamName: string;
+		figmaAdminAtlassianUserId: string;
+		authStatus: FigmaTeamAuthStatus;
+		connectInstallationId: string;
+	}) {
+		this.id = params.id;
+		this.webhookId = params.webhookId;
+		this.webhookPasscode = params.webhookPasscode;
+		this.teamId = params.teamId;
+		this.teamName = params.teamName;
+		this.figmaAdminAtlassianUserId = params.figmaAdminAtlassianUserId;
+		this.authStatus = params.authStatus;
+		this.connectInstallationId = params.connectInstallationId;
+
 		this._adminInfo = {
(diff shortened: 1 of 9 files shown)


src/usecases/remove-figma-team-use-case.test.ts
```diff
@@ -21,10 +21,10 @@ describe('removeFigmaTeamUseCase', () => {
 		expect(
 			figmaTeamRepository.getByTeamIdAndConnectInstallationId,
 		).toBeCalledWith(figmaTeam.teamId, figmaTeam.connectInstallationId);
-		expect(figmaService.tryDeleteWebhook).toBeCalledWith(figmaTeam.webhookId, {
-			atlassianUserId: figmaTeam.figmaAdminAtlassianUserId,
-			connectInstallationId: figmaTeam.connectInstallationId,
-		});
+		expect(figmaService.tryDeleteWebhook).toBeCalledWith(
```
akostevich-atlassian, line 24: Looks so much better! Thank you for the great suggestion, Justin!

src/domain/entities/figma-team.ts
```diff
@@ -1,20 +1,35 @@
...
-	readonly teamId: string;
-	readonly teamName: string;
-	readonly figmaAdminAtlassianUserId: string;
-	readonly authStatus: FigmaTeamAuthStatus;
-	readonly connectInstallationId: string;
-};
+export class FigmaTeam {
+	private readonly _adminInfo: ConnectUserInfo;
+
+	constructor(
```
jv3nk, line 11: /nit given that there's a lot of string parameters in this constructor, it might be better to accept an object instead to avoid accidentally passing params in the wrong order:

```
export class FigmaTeam {
	private readonly _adminInfo: ConnectUserInfo;
	public readonly id: string;
  ...

  	constructor(params: {
		id: string;
    ...
	}) {
		this._adminInfo = {
			atlassianUserId: params.figmaAdminAtlassianUserId,
			connectInstallationId: params.connectInstallationId,
		};
		this.id = params.id;
    ...
	}
```

it is a bit more verbose though, so I'm happy to leave this as is if you don't think it's worth it

jv3nk: If you did make this change, you could also update `FigmaTeamCreateParams` so it doesn't need to be updated if we ever add more getter properties:

```typescript
export type FigmaTeamCreateParams = Omit<
	ConstructorParameters<typeof FigmaTeam>[0],
	'id'
>;
```

jv3nk (approved):

src/domain/entities/figma-team.ts
akostevich-atlassian: That's a very good point, Justin! I totally agree that it gets very easy to introduce a bug by accidentally swapping parameters. I've update the `FigmaTeam` constructor to accepts `params`.

With regards to `FigmaTeamCreateParams`, frankly speaking, I'm leaning towards explicitly listing the required fields for the following reasons:

- It is the simplest and easy to maintain solution.
- As you pointed out, `Omit<FigmaTeam, 'id'|'adminInfo'>;` requires updating the type when a new property is added to `FigmaTeam`.
- `Omit<ConstructorParameters<typeof FigmaTeam>[0], 'id'>` is a quite advance technique that can make the code more difficult to understand. Moreover, it can be subject to change as well, since there is no _direct_ relationship between `FigmaTeam` constructor parameters and `FigmaTeamCreateParams`. They are the same now but they don't have to be the same.

Please, let me know if you have any concerns, Justin.

jv3nk: 👍 makes sense!

PR #90, opened 2023-10-02
NONE Reorganise teams routes and rename use cases

jv3nk:
This PR reorganises the endpoints for managing `FigmaTeam`s.

**OLD**:
- `GET /teams/list` - list teams
- `POST /teams/configure` - connect team
- `DELETE /teams/configure?teamId=:teamId` - disconnect team

**NEW**:
- `GET /teams` - list teams
- `POST /teams/:teamId/connect` - connect team
- `DELETE /teams/:teamId/disconnect` - disconnect team

I've also renamed the `configure-figma-teams` and `remove-figma-team` use cases to `(connect|disconnect)-figma-team` to better match the language used in the config UI.


src/web/routes/teams/integration.test.ts
```diff
@@ -34,11 +34,104 @@ import {
 const figmaTeamSummaryComparer = (a: FigmaTeamSummary, b: FigmaTeamSummary) =>
 	a.teamId.localeCompare(b.teamId);
 
-const TEAMS_CONFIGURE_ENDPOINT = '/teams/configure';
-const TEAMS_LIST_ENDPOINT = '/teams/list';
+const TEAMS_ENDPOINT = '/teams';
 
 describe('/teams', () => {
-	describe('POST /configure', () => {
+	describe('GET', () => {
```
jv3nk, line 40: Just moved this block up to match the route definition order in `teams-router.ts`

akostevich-atlassian (approved):

src/web/routes/teams/types.ts
```diff
@@ -5,50 +5,53 @@ import type {
 	FigmaTeamSummary,
 } from '../../../domain/entities';
 
-export type ConfigureFigmaTeamRequestBody = {
+export type ConnectFigmaTeamRequestBody = {
```
akostevich-atlassian, line 8: **suggestion:** Currently, the name of request/response types do not match the names of the endpoints.

I am wondering, maybe, the idea to use `PUT /teams/:teamId/connect` and `PUT /teams/:teamId/disconnect` as endpoints was not too bad. What do you think, @jv3nk ?

jv3nk: Good point, I'll think about it overnight and come back to this in the morning. I'm not 100% sure about using `PUT`, since we'd be creating a new webhook subscription for each call, so I'm not sure if I'd consider it to be fully idempotent.

akostevich-atlassian (approved):

src/domain/entities/figma-team.ts
```diff
@@ -6,7 +6,7 @@ export enum FigmaTeamAuthStatus {
 }
 
 export class FigmaTeam {
-	private readonly _adminInfo: ConnectUserInfo;
+	readonly #adminInfo: ConnectUserInfo;
```
jv3nk, line 9: We made this change to prevent `_adminInfo` from being included when serialising to JSON, which was causing some integration tests to fail.

PR #91, opened 2023-10-02
MDTZ-1007 Overwrite 'attached-design-url-v2' issue property

jv3nk:
When we write the `attached-design-url-v2` issue property, we check the existing value before updating, as we must include previously associated URLs in the new value. Since anyone with issue access can write any issue property, we also check that the issue property value is in the expected format.

This PR reworks how we update the `attached-design-url-v2` issue property. Instead of throwing an error when the value is in an unexpected format, we overwrite it with the new issue property value instead.

### DIFF
--- a/src/infrastructure/jira/jira-service.ts
+++ b/src/infrastructure/jira/jira-service.ts
@@ -14,7 +14,7 @@ import type {
 	JiraIssue,
 } from '../../domain/entities';
 import { AtlassianAssociation } from '../../domain/entities';
-import { assertSchema } from '../ajv';
+import { assertSchema, SchemaValidationError } from '../ajv';
 import { getLogger } from '../logger';
 
 type SubmitDesignParams = {
@@ -132,6 +132,21 @@ class JiraService {
 		{ url, displayName }: AtlassianDesign,
 		connectInstallation: ConnectInstallation,
 	): Promise<void> => {
+		const setIssueProperty = async (value: string): Promise<void> => {
+			return jiraClient.setIssueProperty(
+				issueIdOrKey,
+				propertyKeys.ATTACHED_DESIGN_URL_V2,
+				value,
+				connectInstallation,
+			);
+		};
+
+		const newAttachedDesignUrlIssuePropertyValue: AttachedDesignUrlV2IssuePropertyValue =
+			{
+				url,
+				name: displayName,
+			};
+
 		let response: GetIssuePropertyResponse;
 		try {
 			response = await jiraClient.getIssueProperty(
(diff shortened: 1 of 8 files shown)


src/infrastructure/jira/jira-service.ts
```diff
@@ -132,6 +132,21 @@ class JiraService {
 		{ url, displayName }: AtlassianDesign,
 		connectInstallation: ConnectInstallation,
 	): Promise<void> => {
+		const setIssueProperty = async (value: string): Promise<void> => {
```
akostevich-atlassian, line 135: **suggestion:** The logic of this property update is quite complicated, and the implementation naturally gets complicated as well. It seems we could use a few tricks to make it shorter, more linear and easier to understand:

- Combine paths when the property is missing and when it has an an unexpected value.
- Use an early return when a property already contains a target URL.

Here is how it could look like (needs to be tested :)):

```ts
	updateAttachedDesignUrlV2IssueProperty = async (
		issueIdOrKey: string,
		{ url, displayName }: AtlassianDesign,
	): Promise<void> => {
		let newValue: AttachedDesignUrlV2IssuePropertyValue[];

		try {
			const response = await jiraClient.getIssueProperty(
				issueIdOrKey,
				propertyKeys.ATTACHED_DESIGN_URL_V2,
				connectInstallation,
			);

			const currentValue = JSON.parse(ensureString(response.value)) as unknown;

			assertSchema<AttachedDesignUrlV2IssuePropertyValue[]>(
				currentValue,
				ATTACHED_DESIGN_URL_V2_VALUE_SCHEMA,
			);

			if (currentValue.some((value) => value.url === url)) return;

			newValue = [...currentValue, { url, name: displayName }];
		} catch (error: unknown) {
			if (
				error instanceof JiraClientNotFoundError ||
				error instanceof SchemaValidationError ||
				error instanceof TypeError ||
				error instanceof SyntaxError
			) {
				newValue = [];
			} else {
				throw error;
			}
		}

		await jiraClient.setIssueProperty(
			issueIdOrKey,
			propertyKeys.ATTACHED_DESIGN_URL_V2,
			this.superStringify(newValue),
			connectInstallation,
		);
	};
```

We could simplify this even further if we introduce an additional utility function for parsing JSON of the expected schema (e.g., `parseJsonOfSchema`):

```ts
	updateAttachedDesignUrlV2IssueProperty = async (
		issueIdOrKey: string,
		{ url, displayName }: AtlassianDesign,
	): Promise<void> => {
		let newValue: AttachedDesignUrlV2IssuePropertyValue[];

		try {
			const response = await jiraClient.getIssueProperty(
				issueIdOrKey,
				propertyKeys.ATTACHED_DESIGN_URL_V2,
				connectInstallation,
			);

			const currentValue = parseJsonOfSchema< // A new utility function that returns a parsed object or throw `SchemaValidationError`.
				AttachedDesignUrlV2IssuePropertyValue[]
			>(response.value, ATTACHED_DESIGN_URL_V2_VALUE_SCHEMA);

			if (currentValue.some((value) => value.url === url)) return;

			newValue = [...currentValue, { url, name: displayName }];
		} catch (error: unknown) {
			if (
				error instanceof JiraClientNotFoundError ||
				error instanceof SchemaValidationError
			) {
				newValue = [];
			} else {
				throw error;
			}
		}

		await jiraClient.setIssueProperty(
			issueIdOrKey,
			propertyKeys.ATTACHED_DESIGN_URL_V2,
			this.superStringify(newValue),
			connectInstallation,
		);
	};
```

akostevich-atlassian (changes requested):

src/infrastructure/jira/jira-service.ts
jv3nk: Great suggestion, looks much easier to follow 🙂. I'll add the parsing util and refactor.

akostevich-atlassian (approved):

PR #95, opened 2023-10-05
MDTZ-1030: Add the skeleton of the "Configure" page

akostevich-atlassian:
## Changes

- **[Improvement]:** Renames a path for static content from `/public` to `/static` to avoid confusion with public endpoints.
- **[Improvement]:** Groups static assets into folders.
- **[Feature]:** Adds the skeleton of the "Configure" page with an example of calling `/teams` API.

image

image


src/atlassian-connect.ts
```diff
@@ -71,16 +71,45 @@ export const connectAppDescriptor = {
...
+				value: 'Get Started',
+			},
+			url: '/static/admin',
+			conditions: [{ condition: 'user_is_admin' }],
+		},
+		webSections: [
+			{
+				key: 'figma-addon-admin-section',
+				name: {
+					value: 'Figma for Jira',
```
jv3nk, line 89: /nit Would be good to use a `const` or something for both this value and `connectAppDescriptor.name` so they always match.

jv3nk (approved):

src/atlassian-connect.ts
akostevich-atlassian: Good point. Done!

PR #98, opened 2023-10-11
MDTZ-1057 Change associateEntity response to return flattened design …

stannnous:
Change associateEntity response to return flattened design entity.

Before:
`
{
  "design": {
    "id": ...
  }
}
`

After:
`
{
  "id": ...
}
`

### DIFF
--- a/src/web/routes/entities/types.ts
+++ b/src/web/routes/entities/types.ts
@@ -9,7 +9,7 @@ import type {
 	DisassociateEntityUseCaseParams,
 } from '../../../usecases';
 
-export type EntitiesResponseBody = AtlassianDesign | string;
+export type EntitiesResponseBody = AtlassianDesign;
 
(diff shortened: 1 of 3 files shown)


src/web/routes/entities/types.ts
```diff
@@ -9,9 +9,7 @@ import type {
 	DisassociateEntityUseCaseParams,
 } from '../../../usecases';
 
-export type EntitiesResponseBody =
-	| { readonly design: AtlassianDesign }
-	| string;
+export type EntitiesResponseBody = AtlassianDesign | string;
```
akostevich-atlassian, line 12: **question:** @stannnous , please, could you remove ` | string` from the type declaration? It seems it is a legacy artifact -- the endpoind should always return `AtlassianDesign` in case of success.

akostevich-atlassian (approved):

PR #103, opened 2023-10-16
MDTZ-1064: Write to ingested design issue property on associate

achandra-atlas:
On association, write the URL of the ingested design to the `ingested-design-urls` issue property, in order to keep track of previously backfilled designs.


jvrankul23 (approved):

src/infrastructure/jira/jira-service.ts
```diff
@@ -180,6 +176,32 @@ class JiraService {
...
+		connectInstallation: ConnectInstallation,
+	): Promise<void> => {
+		const storedValue =
+			await this.getStoredIssuePropertyValue<IngestedDesignUrlIssuePropertyValue>(
+				issueIdOrKey,
+				propertyKeys.INGESTED_DESIGN_URLS,
+				connectInstallation,
+				INGESTED_DESIGN_URL_VALUE_SCHEMA,
+			);
+		if (storedValue && storedValue.some((value) => value === url)) {
```
akostevich-atlassian, line 191: **nitpick:** Consider using the `?` operator to simplify the expression:

```ts
if (storedValue?.some((value) => value === url)) {
  // ...
}
```

The same solution could be applied to the similar code in `updateAttachedDesignUrlV2IssueProperty`.

src/infrastructure/jira/jira-service.ts
```diff
@@ -291,6 +313,31 @@ class JiraService {
 		}
 	};
 
+	private async getStoredIssuePropertyValue<T>(
```
akostevich-atlassian, line 316: **nitpick:** Consider renaming the method to `getIssuePropertyJsonValue` to highlight that it is specific to JSON-encoded values.

For example, while `updateAttachedDesignUrlV2IssueProperty` and `updateIngestedDesignsIssueProperty` read/write JSON values (so, it can use this function),  `setAttachedDesignUrlInIssuePropertiesIfMissing` reads/writes a plain string (so, it cannot use this function).

src/infrastructure/jira/jira-service.ts
```diff
@@ -29,9 +33,12 @@ export type AttachedDesignUrlV2IssuePropertyValue = {
 	readonly name: string;
 };
 
+export type IngestedDesignUrlIssuePropertyValue = string;
+
 export const propertyKeys = {
 	ATTACHED_DESIGN_URL: 'attached-design-url',
 	ATTACHED_DESIGN_URL_V2: 'attached-design-url-v2',
+	INGESTED_DESIGN_URLS: 'ingested-design-urls',
```
akostevich-atlassian, line 41: **suggestion (non-blocking):** Taking into account that issue properties are global, consider adding an app-specifc prefix to minimize the risk of name collisions with other apps (e.g., `figma-for-jira:ingested-design-urls`).

akostevich-atlassian (approved):

src/infrastructure/jira/jira-service.ts
```diff
@@ -95,6 +102,11 @@ class JiraService {
 				design,
 				connectInstallation,
 			),
+			this.updateIngestedDesignsIssueProperty(
```
stannnous, line 105: @jvrankul23 @achandra-atlas can we discuss this decision again? Having reviewed it, I wonder if comparing the relationships if the better option so that the lazy backfill experience is not dependant on the app doing the right thing.

PR #106, opened 2023-10-19
NONE: Add context-qsh as a valid qsh for JWT verification

Roystbeef:
Saw that the JWT token was failing when I tried to use the one I got via `AP.context.getToken()`. After reading through the docs, I believe this is okay and expected that the `qsh` in this case will always be `context-qsh`. So added it as an acceptable value

**Changes:**
- **fix:** Fixes the JWT token failing verification when accessed from the configure page

**Test Plan**
- jest

### DIFF
--- a/src/infrastructure/jira/inbound-auth/jira-jwt-utils.test.ts
+++ b/src/infrastructure/jira/inbound-auth/jira-jwt-utils.test.ts
@@ -1,0 +1,54 @@
+import { createQueryStringHash } from 'atlassian-jwt';
+
+import { verifyQshClaimBoundToUrl } from './jira-jwt-utils';
+
+import { UnauthorizedError } from '../../../web/middleware/errors';
+
+const NOW = Date.now();
+
+describe('verifyQshClaimBoundToUrl', () => {
+	beforeEach(() => {
+		jest.useFakeTimers().setSystemTime(NOW);
+	});
+
+	afterEach(() => {
+		jest.useRealTimers();
+	});
+
+	it('throws on invalid qsh claims', () => {
+		expect(() =>
+			verifyQshClaimBoundToUrl(
+				{ qsh: 'invalid-qsh' },
+				{
+					method: 'GET',
+					pathname: '/foo',
+					query: {},
+				},
+			),
+		).toThrowError(UnauthorizedError);
+	});
+
+	it('passes on valid qsh claims', () => {
+		const request = {
+			method: 'GET',
+			pathname: '/foo',
+			query: {},
+		};
+
+		const qsh = createQueryStringHash(request, false);
+		expect(() => verifyQshClaimBoundToUrl({ qsh }, request)).not.toThrowError();
+	});
+
+	it('passes when the qsh is context-qsh', () => {
+		expect(() =>
+			verifyQshClaimBoundToUrl(
+				{ qsh: 'context-qsh' },
+				{
+					method: 'GET',
+					pathname: '/foo',
+					query: {},
+				},
+			),
+		).not.toThrowError(UnauthorizedError);
+	});
+});
--- a/src/infrastructure/jira/inbound-auth/jira-jwt-utils.ts
+++ b/src/infrastructure/jira/inbound-auth/jira-jwt-utils.ts
@@ -17,7 +17,7 @@ export const verifyQshClaimBoundToUrl = (
 		query?: Record<string, unknown>;
 	},
 ) => {
-	if (qsh !== createQueryStringHash(request, false)) {
+	if (qsh !== createQueryStringHash(request, false) && qsh !== 'context-qsh') {
 		throw new UnauthorizedError('The token contains an invalid `qsh` claim.');
 	}
 };


akostevich-atlassian:
**question:** I apologise, Roy, for this issue: it seem I've missed something in my implementation.

Overall, there a few types of Jira JWT tokens, and they have different requirements to the `qsh` verification (please, see [here](https://developer.atlassian.com/cloud/jira/platform/understanding-jwt-for-connect-apps/) for more detail)):

> Asymmetric JWT token
> 
> For install and uninstall lifecycle callback events, host product will issue **an asymmetrically signed JWT token** which can be used to validate that the request was made from Atlassian.
> 
> Symmetric JWT token (Using shared secret)
> 
>  - **An iframe or server-to-server JWT**, which includes a computed [query string hash](https://developer.atlassian.com/cloud/jira/platform/understanding-jwt-for-connect-apps/#qsh).
>  - **A context JWT** which is not tied to a particular URL. This will include a fixed `qsh` claim.

Therefore, it is important to make sure that we isolate the change to a specific use case.

If I understand correctly, you try to call one of the `/teams` endpoints from the embed configuration page. Embed pages use a context JWT token (obtained using `AP.context.getToken()`), which, as you pointed out, use a fixed `qsh` claims.

Currently, the `/teams` endpoints are protected by `jiraContextSymmetricJwtAuthMiddleware`, which already verifies that the `qsh` claim is equal to `context-qsh`:

- [teamsRouter](https://github.com/atlassian-labs/figma-for-jira/blob/1e03d649818d8b57bedb34287afe9ddfa57155d8/src/web/routes/teams/teams-router.ts#L27)
- [JiraContextSymmetricJwtTokenVerifier](https://github.com/atlassian-labs/figma-for-jira/blob/1e03d649818d8b57bedb34287afe9ddfa57155d8/src/infrastructure/jira/inbound-auth/jira-context-symmetric-jwt-token-verifier.ts#L35)

Therefore, from my understanding, the calls the `/teams` endpoints should work with the existing implementation. 

Please, could you let me know with which specific endpoint you experience the issue?

Roystbeef:
@akostevich-atlassian I ran into this with the `/auth/checkAuth` endpoint! Is the fix here then to just update that (and other endpoints that similarly fail this check on the admin page) to use `jiraContextSymmetricJwtAuthMiddleware`?

akostevich-atlassian:
> @akostevich-atlassian I ran into this with the `/auth/checkAuth` endpoint! Is the fix here then to just update that (and other endpoints that similarly fail this check on the admin page) to use `jiraContextSymmetricJwtAuthMiddleware`?

@Roystbeef, thank you for the explanation! If I understand correctly, you are trying to call  the `/auth/checkAuth` endpoint from the configuration pages. This endpoint was implemented for and is currently used by the Jira backend -- therefore, it currently expects a "server-to-server JWT" instead of a "context JWT". 

It seems that we need to update the implementation, so `/auth/checkAuth` can be used in 2 different use cases:

1. When it is called by the Jira backend, which uses a server-to-server JWT.
2. When it is called by the configuration pages, which use a context JWT.

I will be working on making this change to unblock you and will create a PR shortly.

@Roystbeef, you also mentioned "and other endpoints". Are there any other endpoints required by the configuration pages in addition to the following?

- GET `/auth/checkAuth`
- GET `/teams`
- POST `/teams/:id/connect`
- POST `/teams/:id/disconnect`

akostevich-atlassian:
@Roystbeef , here is the PR that addresses the issue by adding a new `admin/auth/checkAuth` endpoint:

https://github.com/atlassian-labs/figma-for-jira/pull/112

While the PR is in review, please, consider using one of the following workarounds in case you are blocked by this issue:

1. Temporarily replace `jiraServerSymmetricJwtAuthMiddleware` with `jiraContextSymmetricJwtAuthMiddleware` [here](https://github.com/atlassian-labs/figma-for-jira/blob/4bcc5d2132902d8e0c597f24d1e59208f6102c73/src/web/routes/auth/auth-router.ts#L14C16-L14C52) in your local branch.
2. Use the branch for https://github.com/atlassian-labs/figma-for-jira/pull/112 (`NONE-admin-endpoints`).

We will try to get it reviewed and merged asap. Sorry for the inconvenience, Roy, and thank you for sharing this!

Roystbeef:
all good! thanks for following up on this so quickly!

Roystbeef:
> @Roystbeef, you also mentioned "and other endpoints". Are there any other endpoints required by the configuration pages in addition to the following?

@akostevich-atlassian  these are the only ones! I will need a route that can get the email of the figma account that is signed in. We could append this data to the existing `checkAuth` route since the internals of `checkAuth` call `figmaClient.me()` which would return this information. But if we choose to create a new route, that one will need to support the context JWT as well

stannnous:
> > @Roystbeef, you also mentioned "and other endpoints". Are there any other endpoints required by the configuration pages in addition to the following?
> 
> @akostevich-atlassian these are the only ones! I will need a route that can get the email of the figma account that is signed in. We could append this data to the existing `checkAuth` route since the internals of `checkAuth` call `figmaClient.me()` which would return this information. But if we choose to create a new route, that one will need to support the context JWT as well

@Roystbeef Please keep in mind that the `checkAuth` API is also called by Jira and must conform to the contract. Additional fields will be ignored by Jira but please consider this before you modify the schema. For this reason it may be preferable to modify the route used by the admin page but leave the existing `checkAuth` API unchanged.

PR #112, opened 2023-10-23
NONE: Add `/admin/auth/checkAuth` endpoint. Group admin endpoints under `/admin`.

akostevich-atlassian:
The PR addresses an issue with the `checkAuth` endpoint not being accessible from the admin UI. Please, see https://github.com/atlassian-labs/figma-for-jira/pull/106#issuecomment-1774339465 for more detail.

## Changes

- **feat:** Adds a new `/admin/auth/checkAuth` endpoint that can be used to check an admin's auth status from the admin UI.
- **refactor:** Groups all admin-related endpoints under the `/admin` path:
  - `GET /admin/auth/checkAuth`
  - `GET /admin/teams` (before, `GET /teams`)
  - `POST /admin/teams/:teamId/connect` (before, `POST /teams/:teamId/connect`)
  - `POST /admin/teams/:teamId/disconnect` (before, `POST /teams/:teamId/disconnect`)
- **refactor:** Mounts the routes for admin frontend resources under `static/admin` to distinguish the API and static resource routes.

## Remarks

The PR adds a new `/admin/auth/checkAuth` endpoint instead of updating the existing `/admin/auth/checkAuth` endpoint to fit different use cases. While it is tempting to combine these endpoints into 1, there are differences between them that could add a significant complexity into the implementation and testing:

|                        | `/auth/checkAuth`                                                   | `/admin/auth/checkAuth`                                |
|------------------------|---------------------------------------------------------------------|--------------------------------------------------------|
| Purpose                | Implements the standard action required for design integration.     | Implement a custom endpoint for the app UI.            |
| Contact                | Strict (must conform to the Jira contract)                   | Flexible (driven by the app UI needs)                 |
| Authentication         | Server-to-server symmetric JWT. Includes the URL-bound `qsh` claim. | Context symmetric JWT. Includes the fixed `qsh` claim. |
| How is user ID passed? | As `userId` query parameter.                                        | As the `sub` claim in JWT token.                       |

Therefore, the PR reduces the complexity of endpoint implementation at the cost of minor code duplication.

### DIFF
--- a/src/web/routes/admin/auth/auth-router.ts
+++ b/src/web/routes/admin/auth/auth-router.ts
@@ -1,0 +1,40 @@
+import type { NextFunction } from 'express';
+import { Router } from 'express';
+
+import type { CheckAuthRequest, CheckAuthResponse } from './types';
+
+import { figmaAuthService } from '../../../../infrastructure/figma';
+import { checkUserFigmaAuthUseCase } from '../../../../usecases';
+
+export const authRouter = Router();
+
+/**
+ * Checks whether the given Atlassian admin is authorized to call Figma API.
+ */
+authRouter.get(
+	['/checkAuth'],
+	function (req: CheckAuthRequest, res: CheckAuthResponse, next: NextFunction) {
+		const { connectInstallation, atlassianUserId } = res.locals;
+
+		checkUserFigmaAuthUseCase
+			.execute(atlassianUserId, connectInstallation)
+			.then((authorized) => {
+				if (authorized) {
+					return res.send({ authorized });
+				}
+
+				const authorizationEndpoint =
+					figmaAuthService.createOAuth2AuthorizationRequest({
+						atlassianUserId,
+						connectInstallation,
+						redirectEndpoint: `figma/oauth/callback`,
+					});
+
+				return res.send({
+					authorized,
+					grant: { authorizationEndpoint },
+				});
+			})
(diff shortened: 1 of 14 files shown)


src/web/routes/admin/auth/auth-router.ts
```diff
@@ -0,0 +1,40 @@
...
+				const authorizationEndpoint =
+					figmaAuthService.createOAuth2AuthorizationRequest({
+						atlassianUserId,
+						connectInstallation,
+						redirectEndpoint: `figma/oauth/callback`,
+					});
+
+				return res.send({
+					authorized,
+					grant: { authorizationEndpoint },
```
stannnous, line 35: Just noting this comment from @Roystbeef https://github.com/atlassian-labs/figma-for-jira/pull/106#issuecomment-1775418684

and confirming the proposal is to add an email field here without affecting the other `checkAuth` route?

stannnous (approved):

src/web/routes/admin/auth/auth-router.ts
akostevich-atlassian: This is correct. Unlike other endpoints, the `/admin` endpoints are specific to UI and can be changed according to its needs. Therefore, if we need to add a field, we should make this change only for `/admin/auth/checkAuth`.

@Roystbeef , please, let me know if you would like me to take care of this change.

PR #114, opened 2023-10-23
NONE: Support static ngrok urls when making a tunnel

Roystbeef:
Heh, it's annoying to have to update all our callback urls etc and reinstall the connect app when restarting the tunnel. So updated this to support using a static domain we get from ngrok.

Even free users are able to create one at https://dashboard.ngrok.com/cloud-edge/domains although it may be limited to 1

In any case, this PR updates it so that when using the tunnel, it will always resolve to the specified ngrok domain. No more reseting the callback urls and reinstalling the connect app!

### Test plan
- run `npm run start:tunnel`
- Assert that it tunnels to the specified ngrok url in your .env file
- Assert that this does not change when quitting and rerunning the command
- Assert that this url is usable for installing the app, auth, etc

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -39,6 +39,10 @@ JIRA_CONNECT_KEY_SERVER_URL=https://connect-install-keys.atlassian.com
 
 # Auth Token for ngrok: https://dashboard.ngrok.com/get-started/setup
 NGROK_AUTHTOKEN=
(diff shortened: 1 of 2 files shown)


akostevich-atlassian:
**praise:** Fantastic job, Roy! I will not be scared to close my terminal with a running app anymore :)

akostevich-atlassian (approved):

PR #115, opened 2023-10-23
NONE: Add UI and styling for the admin auth page

Roystbeef:
Adds the UI for the auth page

### Screenshots
admin auth page

With the popup
image

### Test plan
- Assert that the auth page matches the designs
- Hover over the "How to check Figma permissions" and assert that the popup loads and matches the designs
- Assert that clicking continue navigates to the figma app auth grant page


admin/src/components/connect-banner.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export function ConnectBanner() {
+	return (
+		<div
+			css={css`
+				display: flex;
+				align-items: center;
+			`}
+		>
+			<div css={css({ padding: 16 })}>
```
jvrankul23, line 12: I should probably know this but does a raw number like `16` work with emotion's `css` prop or do you need to specify the unit of measurement as well?

admin/src/pages/auth/auth-page.tsx
```diff
@@ -11,13 +15,74 @@ type AuthPageProps = {
...
-			</Button>
+				<ConnectBanner />
+				<div
+					css={css`
+						font-weight: 500;
+						font-size: 24px;
+						color: ${token('color.text')};
+					`}
+				>
+					Connect Figma to Jira
```
jvrankul23, line 34: This might be tricky to get a quick solve for, but have we thought about / are we concerned about i18n for non-English speakers?

admin/src/components/connect-banner.tsx
```diff
@@ -0,0 +1,21 @@
...
+				<Image src="/admin/jira-logo.svg" />
```
jvrankul23, line 13: Might be worth also testing this page and some of the assets with Jira's dark mode turned on if you haven't already. `<Image>` has a handy `srcDark` prop that can help if we end up also needing any dark mode assets

admin/src/components/connect-banner.tsx
```diff
@@ -0,0 +1,21 @@
+import Image from '@atlaskit/image';
+import { css } from '@emotion/react';
+
+export function ConnectBanner() {
+	return (
+		<div
```
jvrankul23, line 6: nit: it's nothing major but Atlaskit also now has support for [primitives](https://atlassian.design/components/primitives/overview/)! Might help reduce some of the need for custom styles

jvrankul23 (approved):
Looks good, thanks @Roystbeef! Just left a few minor comments and one around testing in dark mode but I suspect it to be mostly fine considering you've used tokens.

admin/src/components/connect-banner.tsx
Roystbeef: the raw number works here! but only if we're passing an object to `css()`. If we're using tagged template literals, we need to pass in px since the string needs to be valid CSS

admin/src/pages/auth/auth-page.tsx
Roystbeef: agreed that we should definitely look into this! IMO this is worthy of its own project (sans something that exists and would easily plug and play). If you know of any i18n/translation libraries, I'm all ears!

admin/src/components/connect-banner.tsx
Roystbeef: I have not! Will add this to the test plan!

admin/src/components/connect-banner.tsx
Roystbeef: yeah I briefly looked into using Box/Stack/Inline but was thrown off by parts of the API. Maybe worth moving to in the future? Heh, I originally started with tailwind but didn't want to spend time debugging it clashing with the CSS reset we get from Atlassian.

jvrankul23: Makes sense, it is still in beta so no stress with this! Tailwind is also my go to <3

admin/src/components/connect-banner.tsx
Roystbeef: gonna look into dark mode as a separate action item here. I'm noticing that while the app does support natively support dark mode, it inherits from the browser rather than from the atlassian context. So the rest of Jira may still be in light mode and the configure page is rendering in dark mode. Gonna look into whether there's a connect app JS api for determining this and go from there

PR #117, opened 2023-10-24
MDTZ-1081: Fixed string validation error for ingested-design-urls issue property

achandra-atlas:
For the ` ingested-design-urls` issue property schema validation was failing as the value was not a string. Updated the `parseJsonOfSchema` function to handle this

### DIFF
--- a/src/infrastructure/ajv.ts
+++ b/src/infrastructure/ajv.ts
@@ -1,8 +1,6 @@
 import type { ErrorObject, JSONSchemaType, ValidateFunction } from 'ajv';
 import Ajv from 'ajv';
 
-import { ensureString } from '../common/string-utils';
-
 export const ajv = new Ajv({ allowUnionTypes: true });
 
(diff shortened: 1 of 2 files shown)


jvrankul23 (approved):

src/infrastructure/ajv.ts
```diff
@@ -39,9 +37,11 @@ export function parseJsonOfSchema<T>(
 	schema: JSONSchemaTypeWithId<T>,
 ): T {
 	try {
-		const parsed: unknown = JSON.parse(ensureString(value));
-		assertSchema(parsed, schema);
-		return parsed;
+		if (typeof value === 'string') {
```
akostevich-atlassian, line 40: **nipick:** Consider using  the existing `isString` function for a type assertion:

```ts
if (isString(value)) {
  value = JSON.parse(value);
}
```

akostevich-atlassian (approved):

PR #118, opened 2023-10-25
NONE: Add UI and styling for the admin connect team flow

Roystbeef:
### Describe
Adding the screen for connecting a figma team. Upon success, it will render the success screen. Upon error, right now it will show the permissions error. We should update the `/admin/teams/:team_id/connect` route to return a 403 on when it fails to connect the team due to permissions errors and 500 for other errors when appropriate.

### Screenshots:
Connecting a team:
image

When inputting an invalid URL:
image

When there's an error:
image

Success:
image

### Notable changes:
- updated `/admin/teams/:team_id/connect` to return a teamSummary in the success case since we want to render the team name of the team we just connected in the success message

### Things to flag:
- I'm not sure if I need to be added to a feature flag for this to work but I was getting 415 errors with `await jiraService.setConfigurationStateInAppProperties()` which was recently added to `connect-figma-team-use-case.ts`. I just commented that part out when testing this flow

### Test Plan:
- open the configure page
- try to connect a random url
- see that it fails inline
- try to connect a figma team url you don't have admin permissions for
- see that it fails with the permissions section message
- try to connect a figma team url you do have the permissions for
- see that the success screen is rendered upon completion
- see that you can close this page with the `x`, `exit setup`, and `see connected screens`


admin/src/components/link.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+import { openInBrowser } from '../utils';
+
+type LinkProps = {
+	href?: string;
+	onClick?: () => void;
+	children: React.ReactNode;
+};
+
+export function Link({ href, onClick, children }: LinkProps) {
```
jvrankul23, line 11: I might just be blind but I can't see this component being used anywhere? Potentially for link buttons we could use Atlaskits' [link button](https://atlassian.design/components/button/examples#link), or `subtle-link` depending on the use case.

admin/src/components/page.tsx
```diff
@@ -1,3 +1,5 @@
+import Button from '@atlaskit/button';
```
jvrankul23, line 1: Should this file be renamed to `close-button.tsx`?

admin/src/pages/teams/figma-teams-connector.tsx
```diff
@@ -0,0 +1,176 @@
+import LoadingButton from '@atlaskit/button/loading-button';
+import Form, { ErrorMessage, Field, FormFooter } from '@atlaskit/form';
+import SectionMessage from '@atlaskit/section-message';
+import TextField from '@atlaskit/textfield';
+import { token } from '@atlaskit/tokens';
+import { css } from '@emotion/react';
+import { useMutation, useQueryClient } from '@tanstack/react-query';
```
jvrankul23, line 7: `react-query` ❤️

admin/src/pages/teams/figma-teams-connector.tsx
```diff
@@ -0,0 +1,176 @@
...
+		content = (
+			<ConnectTeamSuccessScreen
+				onClose={onClose}
+				teamName={connectTeamMutation.data.teamName}
+			/>
+		);
+	} else {
+		content = (
+			<div
+				css={css`
```
jvrankul23, line 44: More of a general comment, what do you think about extractin styles as constants into `styled.tsx` files (or just outside the component body, top of the file) and passing them in rather than inlining? It might help with component readability, I found it quite tricky to scan this file at a glance (at least in GitHub)

src/infrastructure/repositories/figma-team-repository.integration.test.ts
```diff
@@ -23,21 +23,42 @@ describe('FigmaTeamRepository', () => {
...
+							connectInstallationId: targetConnectInstallation.id,
+						}),
+					)
+					.then((team) =>
+						figmaTeamRepository.getByTeamIdAndConnectInstallationId(
+							team.teamId,
+							targetConnectInstallation.id,
+						),
+					),
+				figmaTeamRepository
```
jvrankul23, line 50: Do we need the 3rd upsert? Looks like we're only using the first two

jvrankul23 (approved):

admin/src/components/link.tsx
Roystbeef: oof - bad commit splitting on my part. This got added in because I was rendering a link for users to sign into a different account. But I changed that into a TODO to keep this PR small since it required updating the `/admin/checkAuth` route. FWIW it's used in the subsequent PR. I'll delete this and just update the subsequent PR to use link button instead!

admin/src/pages/teams/figma-teams-connector.tsx
```diff
@@ -0,0 +1,176 @@
...
+							color: ${token('color.text')};
+							font-size: 24px;
+							font-weight: 500;
+						`}
+					>
+						Connect Figma to Jira
+					</div>
+				</div>
+				{connectTeamMutation.isError && (
+					<ConnectTeamUnauthorizedSectionMessage />
```
jvrankul23, line 72: While I was testing this earlier I discovered there was an issue with the connect team use case where the issue property write was failing and resulting in a `500` from the app (should now be fixed in this [PR](https://github.com/atlassian-labs/figma-for-jira/pull/125)). It threw me a little because I still saw the unauthorized message in the UI. It can be done in separate PRs but I think we should look at returning specific error codes / messages from some of these use cases so we can display more accurate error messages in the UI

jvrankul23:
LGTM thanks @Roystbeef

admin/src/pages/teams/figma-teams-connector.tsx
Roystbeef: agreed! This is what i was referring to in:
> I'm not sure if I need to be added to a feature flag for this to work but I was getting 415 errors with await jiraService.setConfigurationStateInAppProperties() which was recently added to connect-figma-team-use-case.ts. I just commented that part out when testing this flow

and

> Upon error, right now it will show the permissions error. We should update the /admin/teams/:team_id/connect route to return a 403 on when it fails to connect the team due to permissions errors and 500 for other errors when appropriate.

Roystbeef:
ugh i hate pending comments - thought I replied yesterday lmao

src/infrastructure/repositories/figma-team-repository.integration.test.ts
Roystbeef: lol i just used a vim macro to update all the tests that were failing - we don't technically need this but the symmetry is nice i guess

admin/src/pages/teams/figma-teams-connector.tsx
Roystbeef: maybe switching to use tailwind is the solution here? In general I'm in favor of collocating the style definition w the markup since the 2 are intrinsically related. (eg. flex-grow: 1 only makes sense if its parent has display: flex).

PR #119, opened 2023-10-25
NONE: Add UI and styling for show connected teams

Roystbeef:
This PR adds the ui logic for rendering teams that have already been connected:
image

If the webhook has stopped working, it will render the item with an error and have a separate call to action:
image
image
in the even that there is an error, we render the appropriate error message (eg. the user does not have permissions to reconnect or disconnect the team)

## Test Plan
- connect some teams
- assert that they show up in the teams list
- assert that you can add more by clicking add more
- set the webhook status to be error in prisma studio
- assert that we have the appropriate error handling when this is the case


admin/src/pages/teams/connected-teams-list.tsx
```diff
@@ -0,0 +1,311 @@
...
+	);
+}
+
+type ConnectedTeamProps = {
+	team: FigmaTeamSummary;
+	disconnectTeam: (teamId: string) => void;
+	reconnectTeam: (teamId: string) => void;
+};
+
+function ConnectedTeam({
```
jvrankul23, line 158: nit: this file is pretty chonky, thoughts on pulling this (and potentially the error components) out into their own files?

jvrankul23 (approved):

PR #123, opened 2023-10-25
[tests] Tweak the regex for jest unit tests to explicitly check for `integration.`

elainefigma:
The regex bit (?!.*integration.*) checks that we don't match a test like integration.test.ts.

Our copy of this repo lives under the directory `integrations/figma-for-jira`. Currently, this regex inadvertently matches all tests within integrations/figma-for-jira because the folder name has the string integration

To fix the regex, I added a check to exclude integration. vs. just the string integration.

I checked that `npm run test:unit` runs both in this copy and in our copy.

image

### DIFF
--- a/jest.config.unit.ts
+++ b/jest.config.unit.ts
@@ -3,7 +3,7 @@ import baseConfig from './jest.config';
 
 const config: Config = {
 	...baseConfig,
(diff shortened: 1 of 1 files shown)


jvrankul23:
That's a funky problem.. nice catch!

jvrankul23 (approved):

akostevich-atlassian (approved):

PR #125, opened 2023-10-26
NONE: Resolve the issue with setting Connect app property.

akostevich-atlassian:
## Changes

- **fix:** Update `JiraClient.setAppProperty` to 
  - accept `application/json` from the [Set app property](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-app-properties/#api-rest-atlassian-connect-1-addons-addonkey-properties-propertykey-put) endpoint.
  - explicitly serialize "[Set app property](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-app-properties/#api-rest-atlassian-connect-1-addons-addonkey-properties-propertykey-put)" request to JSON as it expects "a valid, non-empty JSON blob".
- **refactor:** Reduce technical debt by updating `JiraClient.setIssueProperty` by analogue with `JiraClient.setAppProperty`. Replaces `superStringify` (confirmed earlier with @atlasctz) with explicit JSON encoding in `JiraClient`.

### DIFF
--- a/src/infrastructure/jira/jira-client/jira-client.ts
+++ b/src/infrastructure/jira/jira-client/jira-client.ts
@@ -160,7 +160,7 @@ class JiraClient {
 				connectInstallation.baseUrl,
 			);
 
-			await axios.put(url.toString(), value, {
+			await axios.put(url.toString(), JSON.stringify(value), {
 				headers: new AxiosHeaders()
 					.setAuthorization(
 						this.buildAuthorizationHeader(url, 'PUT', connectInstallation),
@@ -194,7 +194,7 @@ class JiraClient {
 		});
 
 	/**
-	 * Sets a connect app property
+	 * Sets a connect app property.
 	 *
 	 * @see https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-app-properties/#api-rest-atlassian-connect-1-addons-addonkey-properties-propertykey-put
 	 */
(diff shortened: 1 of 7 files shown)


src/infrastructure/jira/jira-client/jira-client.ts
```diff
@@ -160,7 +160,7 @@ class JiraClient {
 				connectInstallation.baseUrl,
 			);
 
-			await axios.put(url.toString(), value, {
+			await axios.put(url.toString(), JSON.stringify(value), {
```
stannnous, line 163: @akostevich-atlassian I tried this implementation originally but noticed it encodes the string as "\"CONFIGURED\"". Can you confirm the value that get set in app properties as expected? The spec if still a bit vague on this so you may need to talk to the incredibles team.

stannnous (approved):

src/infrastructure/jira/jira-client/jira-client.ts
akostevich-atlassian: Thank you, Sam, for clarifying this! I've updated implementation according to the contact proposed by the Incredibles team.

jvrankul23: Did we test this with the other issue properties we write to? Specifically the `attached-design-url-v2` property as I know there's some funky double `JSON.stringify` business going on with that and just want to make sure we haven't broken anything.

akostevich-atlassian: Thanks for bringing this to my attention to this, Jon! Yes, the PR does not change the behavior of setting `attached-design-url-v2` -- as before, funky double `JSON.stringify` business is still going on :)

We've researched this with Costa but just haven't had time and a good change to apply this refactoring before.

PR #127, opened 2023-10-26
[scripts] Tweak scripts so they cooperate with the Figma repository

elainefigma:
We have a copy of this repo in the Figma codebase. There are some Figma-specific checks/quirks. This PR tweaks the scripts to fix some of the checks. That way, we can reduce discrepancies between the public and the Figma-internal versions.


tsconfig.json
```diff
@@ -17,7 +17,8 @@
 		"noImplicitThis": false,
 		"esModuleInterop": true,
 		"outDir": "build",
-		"baseUrl": "."
+		"baseUrl": ".",
+		"skipLibCheck": true
```
elainefigma, line 21: When we ran CI on the file, I got some TS errors due to discrepancies between @types/jest and @types/mocha. Skipping the lib check resolves it.

scripts/stop-sandbox.sh
```diff
@@ -1,17 +1,16 @@
-#!/usr/bin/env bash
+#!/bin/bash
```
elainefigma, line 1: Our codebase applies some auto-formatting to bash files. I also had to change the first line to get the script to run.

tsconfig.json
akostevich-atlassian: **question:** @elainefigma , please, could you share the TS errors you see? While I do not see any issues with this change, I cannot reproduce the issue and would like to understand why our build is not deterministic.

Please, also, make sure you run `npm ci` instead `npm install` in CI to ensure that the dependencies installation is deterministic and repeatable:
https://betterprogramming.pub/npm-ci-vs-npm-install-which-should-you-use-in-your-node-js-projects-51e07cb71e26

Possibly, the usage of `npm ci` will resolve the issue without changes in `tsconfig.json` 🤞  

Anyway, please, feel free to merge the PR to unblock your work.

akostevich-atlassian (approved):

tsconfig.json
elainefigma: Good call! Running `npm ci` within the admin/ directory as part of CI resolved this error in our build process. Since we don't need this change, I reverted this bit.

akostevich-atlassian (approved):

PR #132, opened 2023-10-30
NONE: Add link to switch users on admin page

Roystbeef:
This PR adds the following UI in the connect team flow:
image

However, in order to show the email and provide a link to re-initiate the auth flow, we needed to refactor our backend a bit. The changes here are:
- `checkUserFigmaAuthUseCase` -> `currentFigmaUserUseCase` so we're grabbing the relevant details about the logged in figma user (we only keep the email rn since that's all we're showing)
- updating `admin/auth/checkAuth` to always return the authorization endpoint so the user can switch accounts even if they are already logged in
- updating the internals of `/auth/checkAuth` to use `currentFigmaUserUseCase` instead. The response from this endpoint is unchanged

### Test Plan
- open the admin page
- try to connect a team
- see that it displays your logged in email at the bottom
- see that clicking the link takes you to the start of the auth flow


src/usecases/current-figma-user-use-case.ts
```diff
@@ -1,12 +1,12 @@
-import type { ConnectInstallation } from '../domain/entities';
+import type { ConnectInstallation, FigmaUser } from '../domain/entities';
 import { figmaService } from '../infrastructure/figma';
 
-export const checkUserFigmaAuthUseCase = {
+export const currentFigmaUserUseCase = {
```
akostevich-atlassian, line 4: **suggestion:** Please, consider creating a separate use case instead of modifying the existing one. Here is why it can be a good idea:

- Separation will allow to scale 2 business use case independently (the same as the corresponding endpoints) and reduce the risk of breaking one while changing another one. While these use cases look similar, they are are of a different nature (one is driven by admin UI, another -- by Jira) and already have differences.
- It will cause only minor code duplication since the majority of the logic is encapsulated in the service layer. These use cases are very thin.
- It is a common good approach to have a separate use case per endpoint since an endpoint is usually represent an entry point to some business operation.

This being the case, consider the following:

- Keep the existing `/auth/checkAuth` endpoint, which calls `checkUserFigmaAuthUseCase`.
- Rename the existing `/admin/auth/checkAuth` endpoint to `/admin/auth/me`, which calls _new_ `getCurrentFigmaUserUseCase`.

Please, let me know if you have any questions or concerns.

Roystbeef: sgtm!

akostevich-atlassian (approved):

akostevich-atlassian:
**praise:** Really great job, Roy!

PR #136, opened 2023-10-31
NONE: pin hashes for images and actions

Roystbeef:
Images referenced from a Dockerfile FROM should always be pinned to an exact SHA-256 hash. This gives us confidence that builds are repeatable and changes to base images are always obvious in CI. This is essentially Docker's equivalent of a "lockfile". Same applies to gh actions

### Test plan
- assert that docker build still works as expected
- assert that the gh action runs as expected

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -4,10 +4,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3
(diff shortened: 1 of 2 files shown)


akostevich-atlassian (approved):
Interesting! Adding this to my "This week I learned" list :) 

If someone would like to read more, I've found the same recommendation in best practices from Snyk:

- https://snyk.io/blog/10-docker-image-security-best-practices/
- https://snyk.io/blog/best-practices-containerizing-python-docker/#:~:text=The%20official%20python%20Docker%20image%20doesn%E2%80%99t%20contain%20a,run%20the%20process%20with%20a%20least%20privilege%20user.

PR #142, opened 2023-11-01
[config] Rename database environment variables

elainefigma:
EDIT: Moving to draft for now while we discuss this change.

We are setting up the database for the figma-for-jira app. In the future, we want to reuse this database for other integrations if needed. To support this, we configured the environment variable names to better reflect this. This PR updates the application logic to rename the database environment variables.

⚠️ To support this change, you will need to update the values in your .env file:
PG_FIGMA_FOR_JIRA_DB_PORT => PG_INTEGRATIONS_PORT
PG_FIGMA_FOR_JIRA_DB_HOST => PG_INTEGRATIONS_HOST
PG_FIGMA_FOR_JIRA_DB_SCHEMA => PG_INTEGRATIONS_DATABASE
PG_FIGMA_FOR_JIRA_DB_ROLE => PG_INTEGRATIONS_JIRA_USER 
PG_FIGMA_FOR_JIRA_DB_PASSWORD => PG_INTEGRATIONS_JIRA_PASSWORD

You may need to uninstall / reinstall the app for the new environment variables to take effect.

## Testing
- Existing CI checks should pass.

PR #143, opened 2023-11-01
[config] Tweak database env variables in entrypoint

elainefigma:
We are setting up the database for the figma-for-jira app. In the future, we want to reuse this database for other integrations if needed. To support this, we configured the environment variable names to better reflect this. This PR updates the entrypoint logic to use the new set of environment variables.

For backwards compatibility, we'll add an if-else check to see which set of environment variables is set.


klin-figma (approved):

stannnous (approved):

akostevich-atlassian:
**question:** @elainefigma , it would be great if we could unify the configuration and use a single set of environment variables to avoid additional complexity and mismatch among environments.

There are at least 3 possible ways forward:

1. Merge the PR as is. At some point, you could remove the `PG_FIGMA_FOR_JIRA_*` environment variables (currently, they are used by Atlassian's deployment pipeline).
2. Rename the `PG_INTEGRATIONS_*` environment variables to `PG_FIGMA_FOR_JIRA_*` in Figma's deployment pipeline (e.g., in the CloudFormation template).
3. Rename the `PG_FIGMA_FOR_JIRA_*` environment variables to `PG_INTEGRATIONS_*` (as was done in https://github.com/atlassian-labs/figma-for-jira/pull/142). Since Figma is the owner of the app, we are more than happy to adapt to Figma's conventions but we will have to verify the feasibility of this change on our side first. Unfortunately, I was not able to get it working from the first attempt :)

Please, let us know what option seems better for you. We are happy to proceed with any of them.

elainefigma:
@akostevich-atlassian Thanks for feedback! Consolidating the environment variables makes sense to me. I'll merge this PR as-is since it'll unblock our deployment pipeline, and it's backwards-compatible. At a later time, I think we can also move forward with option 3 - rename the environment variables as was done in https://github.com/atlassian-labs/figma-for-jira/pull/142

PR #147, opened 2023-11-02
NONE: render a success message on the auth result page

Roystbeef:
When completing the auth flow - in the case that we don't auto close the window, we want to add a success page telling the user to close their window.

lol tbh, I thought about setting up an entire vite/react app for this - but that felt like overkill for 1 page that should be auto closing the majority of the time. So instead, I just inline the styles and added a tiny bit of JS so we don't show the success message if the auth flow failed.

Screenshot:
finally-central-silkworm_ngrok-free_app_static_auth-result__success_true

### Test plan
- assert the window opener message passing still works by testing against https://github.com/atlassian-labs/figma-for-jira/pull/146
- navigate to `/static/auth-result?success=true` and assert that the success screen shows up
- navigate to `/static/auth-result?success=false` and assert that the screen is blank

### DIFF
--- a/static/auth-result/index.html
+++ b/static/auth-result/index.html
@@ -4,19 +4,178 @@
     <meta charset="utf-8" />
     <title></title>
   </head>
-  <body>
+  <body
+    style="
+      font-family:
+        -apple-system,
+        BlinkMacSystemFont,
+        Segoe UI,
+        Roboto,
+        Oxygen,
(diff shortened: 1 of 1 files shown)


Roystbeef:
closing as this is a dupe of https://github.com/atlassian-labs/figma-for-jira/pull/148

PR #149, opened 2023-11-04
[entrypoint] Add password encoding in the database URL

elainefigma:
This PR updates the entrypoint bash script to URL encode the characters. Prisma expects the database URL to be URL-encoded - [see Prisma docs](https://www.prisma.io/docs/reference/database-reference/connection-urls).


entrypoint.sh
```diff
@@ -1,12 +1,26 @@
 #!/bin/bash
 
-set -e
```
akostevich-atlassian, line 3: **question:** @elainefigma , do we need to remove `set -e`? If I understand correctly, `-e` means " Exit immediately if a command exits with a non-zero status.", which seems like the correct behavior for this script.

entrypoint.sh
```diff
@@ -1,12 +1,26 @@
...
+  echo "$encoded"
+}
 
 # This supports both the old environment variables and new ones specific to the Figma setup.
 if [ -n "$PG_FIGMA_FOR_JIRA_DB_ROLE" ]; then
-  DATABASE_URL=postgresql://${PG_FIGMA_FOR_JIRA_DB_ROLE}:${PG_FIGMA_FOR_JIRA_DB_PASSWORD}@${PG_FIGMA_FOR_JIRA_DB_HOST}:${PG_FIGMA_FOR_JIRA_DB_PORT}/${PG_FIGMA_FOR_JIRA_DB_SCHEMA}
+  DATABASE_URL="postgresql://${PG_FIGMA_FOR_JIRA_DB_ROLE}:${PG_FIGMA_FOR_JIRA_DB_PASSWORD}@${PG_FIGMA_FOR_JIRA_DB_HOST}:${PG_FIGMA_FOR_JIRA_DB_PORT}/${PG_FIGMA_FOR_JIRA_DB_SCHEMA}"
 else
-  DATABASE_URL=postgresql://${PG_INTEGRATIONS_JIRA_USER}:${PG_INTEGRATIONS_JIRA_PASSWORD}@${PG_INTEGRATIONS_HOST}:${PG_INTEGRATIONS_PORT}/${PG_INTEGRATIONS_DATABASE}
+  ENCODED_PASSWORD="$(urlencode "$PG_INTEGRATIONS_JIRA_PASSWORD")"
```
akostevich-atlassian, line 22: **suggestion:** Please, consider encoding other components as well.

According to the [Prisma docs](https://www.prisma.io/docs/reference/database-reference/connection-urls#special-characters):

> For MySQL, PostgreSQL and CockroachDB you must [percentage-encode special characters](https://developer.mozilla.org/en-US/docs/Glossary/percent-encoding) in **any part** of your connection URL - including passwords.

entrypoint.sh
```diff
@@ -1,12 +1,26 @@
...
+    *) encoded+="%"$(printf "%02X" "'$char") ;;
+    esac
+  done
+  echo "$encoded"
+}
 
 # This supports both the old environment variables and new ones specific to the Figma setup.
 if [ -n "$PG_FIGMA_FOR_JIRA_DB_ROLE" ]; then
-  DATABASE_URL=postgresql://${PG_FIGMA_FOR_JIRA_DB_ROLE}:${PG_FIGMA_FOR_JIRA_DB_PASSWORD}@${PG_FIGMA_FOR_JIRA_DB_HOST}:${PG_FIGMA_FOR_JIRA_DB_PORT}/${PG_FIGMA_FOR_JIRA_DB_SCHEMA}
+  DATABASE_URL="postgresql://${PG_FIGMA_FOR_JIRA_DB_ROLE}:${PG_FIGMA_FOR_JIRA_DB_PASSWORD}@${PG_FIGMA_FOR_JIRA_DB_HOST}:${PG_FIGMA_FOR_JIRA_DB_PORT}/${PG_FIGMA_FOR_JIRA_DB_SCHEMA}"
```
akostevich-atlassian, line 20: **question:** @elainefigma , do we need to apply the same change here as well to avoid mismatch between configurations? I will verify the behaviour on our side and can apply a fix if required.

akostevich-atlassian (approved):
**praise:** Oh, what's a great catch! Fantastic job, @elainefigma !

entrypoint.sh
elainefigma: Ah, this is my bad. I put this back.

entrypoint.sh
elainefigma: That suggestion makes sense to me! I think in our case, we don't have any special characters in the other configuration strings, and they're unlikely to change. If there are no special characters, then urlencode is a no-op, so I think this is alright as-is for now.

```
Elaine $ foo="ABC"
Elaine $ bar="$(urlencode $foo)"
Elaine $ echo "$bar"
ABC
```

PR #153, opened 2023-11-08
Add logic to parse starter team URLs

elainefigma:
## Description

Figma has different pricing / plans - [see link](https://www.figma.com/pricing/). For starter teams, we do not support webhooks. In the figma-for-jira configuration page, currently, if you put a starter team URL, then you will see an error like `You do not have admin permissions for this team.`

This PR updates the flow to show a more descriptive error message.

Fixes: https://figlassian.atlassian.net/jira/software/projects/FABB/boards/3?assignee=64139c639d2bc6c90a8a20fd&selectedIssue=FABB-2

## Testing

I wrote a unit test that passes before and after.
image

I tested the changes in the UI.
image


Roystbeef:
Ha, as it turns out - webhooks aren't enabled for starter accounts
image
hence why the request is failing.
Unfortunately this isn't super well documented but @akostevich-atlassian do you have thoughts here RE: how best to parse error responses from Figma's REST API?

This isn't well documented but from testing, I found that this call will return:
```ts
{
  error: true,
  status: 400,
  message: 'Access Denied',
  i18n: null,
  reason: 'Upgrade to professional team to enable webhooks'
}
```
in `error.response?.data`.

Updating `HttpClientError` to something like:
```ts
export class HttpClientError extends CauseAwareError {
  constructor(message: string, cause: AxiosError) {
    super(message, cause);
  }
}
```
could work since we'd be able to access `error.cause.response?.data` but this didn't feel compatible w the other errors that extend `CauseAwareError` eg. `MissingOrInvalidCredentialsFigmaAuthServiceError`. WDYT?

akostevich-atlassian:
Good question, @Roystbeef !

I believe there 2 possible solutions here.

## 1. Use a team URL as an indicator of an account type

Our frontend could use a given URL to determine a Figma Starter account and show an error even before hitting API.

This solution can be the easiest to implement but we need to consider whether a Figma URL structure is reliable indicator of a Figma account type (e.g., are there any other URL formats, is a URL subject to change?). 

## 2. Use API response to show an error

Our frontend could show a more specific error using a response from the `POST admin/teams/:teamId/connect` endpoint.
I suggest to show something like "Cannot connect the team. Please, verify that you have a Figma Professional account and are the admin of the team." when the endpoint returns `HTTP 403 Forbidden`.

Currently, `POST admin/teams/:teamId/connect`, likely, returns `HTTP 500 Internal Server Error` for this use case, so we need to extend error handling for it to consider the new error.

Before jumping in into how to do this, please, let me provide the overview of the error handling design in the app. Overall, the app has several layers of errors: 

- **Base Errors**. They represent errors, which extend a built-in `Error` with additional details and can be used as a base class for _any_ other errors in the app. Currently, there is only one base error, `common/errors/CauseAwareError`, which include information on the original cause of an error (also known as an inner error).
- **External Client Errors.** They represent _app-agnostic_ errors returned from external system (e.g., `HttpClientError`). These errors are thrown by corresponding clients. For example, `FigmaClient` throws `HttpClientError`. 
- **Service Errors.** They represent domain-specifc errors, which we are interested in our business logic (e.g., `MissingOrInvalidCredentialsFigmaAuthServiceError`, `UnauthorizedFigmaServiceError`). There errors are thrown by services. Usually, our services translate relevant *External Client Errors* to *Service Errors* (see an example [here](https://github.com/atlassian-labs/figma-for-jira/blob/52fe0da7d97553db25e2d83ec37aadc27c885495/src/infrastructure/figma/figma-service.ts#L342)) or throw the latter directly.
- **Use Case Errors.** They represent app-specific user-facing errors of a business use case (e.g.,  `ForbiddenByFigmaUseCaseResultError`). Our use cases translate relevant *Service Errors* to *Use Case Error* . There errors are mapped to corresponding HTTP codes and returned to a user.

For our problem, I recommend the following:

1. Add `BadRequestHttpClientError` to [/infrastructure/http-client-errors.ts](https://github.com/atlassian-labs/figma-for-jira/blob/main/src/infrastructure/http-client-errors.ts), which corresponds to `HTTP 400 Bad Request` (we haven't added it before since we haven't been interested in these errors).
2. Extend `HttpClientError` to include response data. A response will be needed to determine whether an error on webhook creation is a Starter account-related error
  ```ts
  export class HttpClientError extends CauseAwareError {
	constructor(message: string, readonly response?: unknown, cause?: unknown) {
		super(message, cause);
	}
  }
  ```
3. Update [translateAxiosError](https://github.com/atlassian-labs/figma-for-jira/blob/main/src/infrastructure/axios-utils.ts#L22) to translate `HTTP 400 Bad Request` to `BadRequestHttpClientError`. Include response `error.response.data` 
  ```ts
  const translateAxiosError = (error: unknown): unknown => {
	if (!isAxiosError(error)) return error;

	switch (error.response?.status) {
		case HttpStatusCode.BadRequest:
			return new BadRequestHttpClientError(error.message, error.response.data, error);
		case HttpStatusCode.Unauthorized:
			return new UnauthorizedHttpClientError(error.message, error.response.data, error);
		// ...
	}
};
```
4. Update [FigmaService.createFileUpdateWebhook](https://github.com/atlassian-labs/figma-for-jira/blob/f809e6c6b8f3c665c0a07c672eb921e2e72d9170/src/infrastructure/figma/figma-service.ts#L211C2-L211C25) to translate `BadRequestHttpClientError` (thrown from `figmaClient.createFileUpdateWebhook`) with `error.response.message === 'Access Denied'` to `UnauthorizedFigmaServiceError` and throw the latter further. You might need to check a type of `error.response` using `validateSchema`.

Considering that [connectFigmaTeamUseCase.execute](https://github.com/atlassian-labs/figma-for-jira/blob/main/src/usecases/connect-figma-team-use-case.ts) already handles `UnauthorizedFigmaServiceError`(🙌 ), the result of these steps should be `POST admin/teams/:teamId/connect` returning `HTTP 403 Forbidden`. Done!

Please, let me know, @Roystbeef , if you would like to discuss this in more detail, need help or have any questions.

**Comment by Roystbeef** [comment]
@akostevich-atlassian this makes sense to me! Thanks for laying this out!

File: admin/src/utils/parseTeamIdFromFigmaUrl.test.ts [hunk 1]
```diff
@@ -0,0 +1,24 @@
+import { parseTeamIdFromFigmaUrl } from './parseTeamIdFromFigmaUrl';
+
+describe('parseTeamIdFromFigmaUrl', () => {
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
elainefigma, line 4: @akostevich-atlassian @Roystbeef I suspect there's something odd with the TS configuration for the admin/ folder. I couldn't figure out what it was, so I'm suppressing the lint checks for the time-being.

admin/src/utils/parseTeamIdFromFigmaUrl.ts
```diff
@@ -18,13 +18,19 @@ export function parseTeamIdFromFigmaUrl(figmaUrl: string): string {
 			throw new FigmaURLError('Invalid Figma URL');
 		}
 		const pathname = url.pathname;
-		const teamIdRegex = /\/files\/\d+\/team\/(\d+)/;
-		const match = pathname.match(teamIdRegex);
-		if (!match || !match[1]) {
-			throw new FigmaURLError('Unable to parse team ID from Figma URL');
+		const orgTeamRegex = /\/files\/\d+\/team\/(\d+)/;
```
elainefigma, line 21: Note: In the Figma UI, when you copy the URL on a paid plan, it matches this URL. However, if you are a starter plan, the default URL format is different. However, the URL does not necessarily indicate the user's plan. For instance, you can reformat a team ID for a starter team to match /files/\d+/team/<starter team ID>.

elainefigma:
@akostevich-atlassian Thank you for the detailed guidance - it was very helpful. I updated the PR to implement the suggestions. I'm OOO for the next week. I'll be back Tue 11/28. @Roystbeef will take point on wrapping this up.

admin/src/pages/teams/figma-teams-connector.tsx
```diff
@@ -46,13 +58,24 @@ export function FigmaTeamConnector({
...
 			return (await connectTeam(teamId)).data;
 		},
 		onSuccess: () => {
+			setShowUnauthorizedError(false);
 			return queryClient.invalidateQueries({ queryKey: ['teams'] });
 		},
+		onError: (error: ConnectTeamsError) => {
+			if (error.response?.data?.detail === WEBHOOK_UPGRADE_ERROR) {
+				setShowUnauthorizedError(false);
+				setValidationError('You need a paid Figma plan to add teams to Jira');
```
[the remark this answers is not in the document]
akostevich-atlassian: @elainefigma , let me ask Jade whether we could show a different message.

admin/src/utils/parseTeamIdFromFigmaUrl.test.ts
akostevich-atlassian: @elainefigma , I've resolved the issue with a temporary workaround.

@Roystbeef , @elainefigma , in order to prevent these tricky issues, I highly recommend separating backend and frontend projects into separate folders and avoid sharing configuration (e.g., for ESLint, Jest, etc.) between them. The frontend and backend projects are very different and might be very difficult to maintain setup that fits both of them.

```
/figma-for-jira
├─ frontend/
│  ├─ src/
│  ├─ eslintrc.cjs
│  ├─ package.json
│  ├─ tsconfig.json
│  ├─ ...
├─ backend/
│  ├─ src/
│  ├─ eslintrc.cjs
│  ├─ package.json
│  ├─ tsconfig.json
│  ├─ ...
├─ docker-compose.yml
├─ package.json
├─ README.md
├─ ...
``` 

What do you think?

src/infrastructure/figma/figma-service.ts
```diff
@@ -353,7 +354,8 @@ export class FigmaService {
 			if (
 				e instanceof MissingOrInvalidCredentialsFigmaAuthServiceError ||
 				e instanceof UnauthorizedHttpClientError ||
-				e instanceof ForbiddenHttpClientError
+				e instanceof ForbiddenHttpClientError ||
+				e instanceof BadRequestHttpClientError
```
akostevich-atlassian, line 358: **issue:** It can be risky to treat any `HTTP 400 Bad Request` as `UnauthorizedFigmaServiceError`. Instead, we should apply this translation only to specific endpoints (where `HTTP 400 Bad Request` really means "Unauthorized"). I will address this issue.

src/usecases/associate-design-use-case.ts
```diff
@@ -85,7 +85,7 @@ export const associateDesignUseCase = {
 			return design;
 		} catch (e) {
 			if (e instanceof UnauthorizedFigmaServiceError) {
-				throw new ForbiddenByFigmaUseCaseResultError({ cause: e });
```
akostevich-atlassian, line 88: **praise:** Nice catch!

src/usecases/connect-figma-team-use-case.ts
```diff
@@ -52,7 +59,11 @@ export const connectFigmaTeamUseCase = {
 			return figmaTeam.toFigmaTeamSummary();
 		} catch (e) {
 			if (e instanceof UnauthorizedFigmaServiceError) {
-				throw new ForbiddenByFigmaUseCaseResultError({ cause: e });
+				if (e.cause instanceof BadRequestHttpClientError) {
```
akostevich-atlassian, line 62: **suggestion:** It could be benefitial to handle `BadRequestHttpClientError` in `FigmaService.createFileUpdateWebhook`, so:

1.  `FigmaService.createFileUpdateWebhook` handles `BadRequestHttpClientError` and throw a more specific `PaidPlanRequiredFigmaServiceError`.
2. `connectFigmaTeamUseCase` handles the very specific `PaidPlanRequiredFigmaServiceError`.

This can bring the following benefits:

- `connectFigmaTeamUseCase` does not know about what errors my _transient_ dependencies (dependencies of my dependencies) throw. Therefore, it handles only errors that are part of contract of its _direct_ dependencies. We try to document errors that are part of the function/method contract: e.g., https://github.com/atlassian-labs/figma-for-jira/blob/3100c153d54723111745fd5829d41d42b35fda30/src/infrastructure/figma/figma-service.ts#L224.
- It allows to handle `HTTP 400` specifically from `POST /v2/webhooks`. Currently, the code catch all `HTTP 400` regardless whether they are thrown from Figma API or Jira API.

I will try to address this to get the PR merged ASAP.

akostevich-atlassian:
@elainefigma , I created a separate PR to address the comments here:
https://github.com/atlassian-labs/figma-for-jira/pull/171

Thank you very much for your great work on this!

admin/src/utils/parseTeamIdFromFigmaUrl.test.ts
Roystbeef: yeah that's a good call

Roystbeef (approved):
the changes in here look good to me! Thank you @elainefigma and @akostevich-atlassian for your work on this!!

PR #154, opened 2023-11-08
[logging] Add the Datadog tracer 

elainefigma:
When we deploy the figma-for-jira app, we use Datadog to track the service metric / logs. This PR adds Datadog to the public repo. In practice, this should be a no-op.

### DIFF
--- a/src/infrastructure/tracer.ts
+++ b/src/infrastructure/tracer.ts
@@ -1,15 +1,17 @@
 import ddTrace from 'dd-trace';
 
+import { getConfig } from '../config/config';
+
 const tracer = ddTrace.init({
-	service: process.env.FIGMA_SERVICE_NAME || '',
+	service: getConfig().tracer.service,
 	profiling: true,
-	env: process.env.DD_ENV || '',
+	env: getConfig().tracer.env,
 	runtimeMetrics: true,
 	logInjection: true,
-	tags: { figma_service: process.env.FIGMA_SERVICE_NAME || '' },
+	tags: { figma_service: getConfig().tracer.service },
 });
 
 tracer.use('express', {
 	enabled: true,
-	service: process.env.FIGMA_SERVICE_NAME || '',
+	service: getConfig().tracer.service,
 });
--- a/src/infrastructure/tracer.ts
+++ b/src/infrastructure/tracer.ts
@@ -1,0 +1,15 @@
+import ddTrace from 'dd-trace';
+
+const tracer = ddTrace.init({
+	service: process.env.FIGMA_SERVICE_NAME || '',
+	profiling: true,
+	env: process.env.DD_ENV || '',
+	runtimeMetrics: true,
+	logInjection: true,
+	tags: { figma_service: process.env.FIGMA_SERVICE_NAME || '' },
+});
+
+tracer.use('express', {
+	enabled: true,
+	service: process.env.FIGMA_SERVICE_NAME || '',
+});
--- a/src/web/routes/router.ts
+++ b/src/web/routes/router.ts
@@ -1,3 +1,7 @@
+// Must come before importing any instrumented module.
+// eslint-disable-next-line import/no-unassigned-import
+import '../../infrastructure/tracer';
+
 import { HttpStatusCode } from 'axios';
 import type { Request, Response } from 'express';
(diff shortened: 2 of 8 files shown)


src/web/routes/router.ts
```diff
@@ -1,3 +1,7 @@
+// Must come before importing any instrumented module.
```
akostevich-atlassian, line 1: **suggestion:** Consider moving this line to `/src/app.ts`, which represents an entry point to the application and is responsible for overall setup.

https://github.com/atlassian-labs/figma-for-jira/blob/341e806b610a5279120dc8faba7e8d1e9c24188a/src/app.ts#L1C1-L19C20

src/infrastructure/tracer.ts
```diff
@@ -0,0 +1,15 @@
+import ddTrace from 'dd-trace';
+
+const tracer = ddTrace.init({
+	service: process.env.FIGMA_SERVICE_NAME || '',
```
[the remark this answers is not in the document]
akostevich-atlassian: **suggestion:** Please, consider adding the new environment variables to `.env.example` and adding comments to help newcomers to set them correctly. If they are optional for local development, we could also mention this there.

src/infrastructure/tracer.ts
```diff
@@ -0,0 +1,15 @@
...
+	profiling: true,
```
akostevich-atlassian, line 5: **question:** @elainefigma , I apologise for not verifying the changes on our environment -- I will do this early tomorrow. Please, could you help me to understand what makes `dd-tracer` to switch between the "no-op" behavior and integration with Datadog?

From what I see in the [docs](https://docs.datadoghq.com/tracing/trace_collection/library_config/nodejs/#instrumentation), Datadog enables many features by default. Sorry if I am missing something: I haven't set up Datadog before.

image

elainefigma: Apologies for the confusion - it's not entirely a no-op. [Here](https://docs.datadoghq.com/tracing/trace_collection/?tab=containers#install-and-configure-the-agent) is a diagram for reference. 

To clarify, in our setup, when we deploy figma-for-jira, we also spin up a Datadog agent. When you add the Datadog tracing library to your code, it instruments the service and sends traces to the Datadog Agent. The Agent then sends the traces to the Datadog backend to be displayed in the UI.

In the dev setup and the Jira setup, there is no Datadog agent. In these environments, we send traces, and nothing happens since the Datadog agent is not set up.

src/infrastructure/tracer.ts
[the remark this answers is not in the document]
elainefigma: Thanks for the thorough review! I've made the suggested updates.

src/infrastructure/tracer.ts
akostevich-atlassian: Got it! Thank you very much for the clarification, @elainefigma -- it all makes sense now!

This being the case, I also recommend disabling `dd-trace` features completely for non-Figma environments. We could achieve this by using the [DD_TRACE_ENABLED](https://docs.datadoghq.com/tracing/trace_collection/library_config/nodejs/#instrumentation) environment variable.

Please, consider adding `DD_TRACE_ENABLED=false` to our `.env.example`:

```
# Optional. These env variables are used to configure Datadog for tracing.
DD_TRACE_ENABLED=false
DD_SERVICE=figma-for-jira
DD_ENV=dev
```

akostevich-atlassian: Just for sharing... It seems that Datadog uses this env variable to choose between the "noop" and proxy-based tracing implementations:

https://github.com/DataDog/dd-trace-js/blob/0999ce6ab2bcde6ab77ad502f03a7c203667050d/packages/dd-trace/src/index.js#L8

akostevich-atlassian (changes requested):
Please, see [this comment](https://github.com/atlassian-labs/figma-for-jira/pull/154/files#r1387317458).

src/infrastructure/tracer.ts
elainefigma: That suggestion makes sense to me - thank you for looking into this!

akostevich-atlassian (approved):

PR #155, opened 2023-11-08
FABB-1: Fix admin page theming

Roystbeef:
Followed the instructions on https://developer.atlassian.com/cloud/jira/platform/connect-theming/ and it works!

Adding it into the `index.html` file directly so as to avoid any flashing. It does need to be run after
```html
<script
  src="https://connect-cdn.atl-paas.net/all.js"
  data-options="sizeToParent:true"
></script>
```
since this sets `window.AP`.

### Test Plan
- Navigate to your personal settings and toggle Jira to be in dark/light mode
- Go to the configure page and assert that the app updates to reflect the theme Jira uses

### DIFF
--- a/admin/index.html
+++ b/admin/index.html
@@ -11,6 +11,9 @@
       src="https://connect-cdn.atl-paas.net/all.js"
       data-options="sizeToParent:true"
     ></script>
+    <script>
+      window.AP.theming.initializeTheming();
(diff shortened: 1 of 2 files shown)


elainefigma (approved):
I tested locally, and this change makes sense to me!

PR #156, opened 2023-11-08
FABB-4: Update switch accounts flow to navigate directly to the accounts picker

Roystbeef:
This PR updates the logic for clicking `Change Figma account` to navigate directly to Figma's account selector page instead of the auth grant screen.

### Test Plan
- Click `Change Figma account`
- Assert that it navigates to the account picker page in Figma
- Assert that this works even when you're only logged into 1 account (screenshot)
image
- Assert that after selecting an account, it will go through the auth grant flow


elainefigma (approved):
I tested locally, and this change makes sense to me!

PR #160, opened 2023-11-09
MDTZ-1175: Remove Figma File name from design URL. Encode URL path parameters in API clients.

akostevich-atlassian:
## Changes

- **feat:** Exclude a Figma File name from the design URL as it is not required and is subject to change. The app still include File name in the URL written to the `attached-design-url` Issue Property for compatibility with "Jira" widget in Figma (which extracts a name from the URL).
- **feat:** Encode dynamic URL path parameters using `encodeURIComponent` when construct URLs (e.g., in API clients).
- **test:** Add unit tests for utility functions responsible for constructing Figma URLs.

### DIFF
--- a/src/infrastructure/figma/transformers/utils.ts
+++ b/src/infrastructure/figma/transformers/utils.ts
@@ -6,7 +6,6 @@ import { getConfig } from '../../../config';
  */
 export const buildDesignUrl = ({
 	fileKey,
-	fileName,
 	nodeId,
 }: {
 	fileKey: string;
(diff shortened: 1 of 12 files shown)


src/infrastructure/figma/transformers/utils.ts
```diff
@@ -23,30 +23,33 @@ export const buildDesignUrl = ({
 };
 
 /**
- * Builds a Live Embed URL to a Figma design given Figma file/node metadata.
- * Inspect URL is used as this is the preferred mode of display in Jira.
- * @see https://www.figma.com/developers/embed
+ * Builds an Inspect Mode URL to a Figma design given Figma file/node metadata.
  */
-export const buildLiveEmbedUrl = ({
```
akostevich-atlassian, line 30: **note:** The PR deletes a Figma File name from URLs and changes the order of function declarations to put dependent functions after their dependencies.

stannnous (approved):

PR #161, opened 2023-11-11
MDTZ-1187 get status from ancestor node if not available in current one

pkovalsky-atlassian:
https://softwareteams.atlassian.net/browse/MDTZ-1187


akostevich-atlassian (approved):
Really great job, Peter!

PR #167, opened 2023-11-14
NONE: Exclude potentially sensitive information from HTTP request/response logs

akostevich-atlassian:
## Changes

- **feat:** Exclude potentially sensitive information from HTTP request/response logs (e.g., HTTP headers, remote address, etc.).

## Remarks

The app will log the following information:

- Request
  - ID
  - HTTP method
  - URL
- Responses
  - HTTP Code
- Response time
- Thrown error (all fields)

An example of a successful `/atlassian-connect.json` request:

image

An example of an unsuccessful `/admin/teams/:teamId/connect` request:

image


elainefigma (approved):
Thanks for making this change! This makes sense to me.

PR #168, opened 2023-11-15
NONE: Improve webhook reliability

akostevich-atlassian:
## Changes

- **fix:** Skips Figma `FILE_UPDATE` event processing when Figma File is not found (an edge case when a Figma File is deleted right after a webhook is triggered) and return `HTTP 200`.
- **fix:** Does not fail Figma `FILE_UPDATE` event processing when a File does not contain some Nodes, previously linked to Jira issues. Skips event processing for these Nodes but processes the remaining. 
- **fix:** Returns `HTTP 200` for Figma `FILE_DELETE` event, which is delivered within the `FILE_UPDATE` subscription. According to Figma [docs](https://www.figma.com/developers/api#webhooks-v2-events), "If you subscribe to FILE_UPDATE, you automatically get these notifications.".

### DIFF
--- a/src/infrastructure/figma/figma-service.test.ts
+++ b/src/infrastructure/figma/figma-service.test.ts
@@ -40,6 +40,7 @@ import {
 } from '../../domain/entities/testing';
 import {
 	ForbiddenHttpClientError,
+	HttpClientError,
 	NotFoundHttpClientError,
 } from '../http-client-errors';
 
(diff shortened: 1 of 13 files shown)


src/infrastructure/figma/figma-service.test.ts
```diff
@@ -176,36 +178,37 @@ describe('FigmaService', () => {
 			});
 		});
 
-		it('should throw when a request to a figma api fails', async () => {
```
stannnous, line 179: Is it valuable to keep the "figma api fails" tests as well?

stannnous (approved):

src/infrastructure/figma/figma-service.test.ts
akostevich-atlassian: Taking into account, that the method does error translation, yes! Returned the test!

PR #169, opened 2023-11-16
FABB-50: Improve compatibility with the "Jira" widget in Figma

akostevich-atlassian:
## Changes

- **fix:** Resolves the issue with the "Jira" widget showing an already unlinked design.
- **fix:** Resolves the issue with the "Jira" widget showing some special symbols in a design name incorrectly.
  image
  image
- **refactor:** Removes unused parameter from Figma design transformation utilities.

### DIFF
--- a/src/infrastructure/figma/transformers/figma-file-meta-transformer.test.ts
+++ b/src/infrastructure/figma/transformers/figma-file-meta-transformer.test.ts
@@ -42,18 +42,9 @@ describe('transformFileMetaToAtlassianDesign', () => {
 		expect(result).toStrictEqual({
 			id: fileKey,
 			displayName: fileMetaResponse.file.name,
-			url: buildDesignUrl({
-				fileKey,
-				fileName: fileMetaResponse.file.name,
-			}),
-			liveEmbedUrl: buildLiveEmbedUrl({
-				fileKey,
-				fileName: fileMetaResponse.file.name,
-			}),
-			inspectUrl: buildInspectUrl({
-				fileKey,
-				fileName: fileMetaResponse.file.name,
-			}),
+			url: buildDesignUrl({ fileKey }),
+			liveEmbedUrl: buildLiveEmbedUrl({ fileKey }),
+			inspectUrl: buildInspectUrl({ fileKey }),
 			status: AtlassianDesignStatus.NONE,
 			type: AtlassianDesignType.FILE,
 			lastUpdated: fileMetaResponse.file.last_touched_at,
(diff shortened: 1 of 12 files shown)


src/infrastructure/figma/transformers/figma-file-meta-transformer.test.ts
```diff
@@ -42,18 +42,9 @@ describe('transformFileMetaToAtlassianDesign', () => {
 		expect(result).toStrictEqual({
 			id: fileKey,
 			displayName: fileMetaResponse.file.name,
-			url: buildDesignUrl({
-				fileKey,
-				fileName: fileMetaResponse.file.name,
```
akostevich-atlassian, line 47: **note:** Just removes `fileName` as it is unused by the `buildDesignUrl` function (likely, the code was merged incorrectly).

src/infrastructure/jira/jira-service.ts
```diff
@@ -298,13 +298,14 @@ class JiraService {
...
 
-			if (storedUrl === design.url) {
-				await jiraClient.deleteIssueProperty(
-					issueIdOrKey,
-					issuePropertyKeys.ATTACHED_DESIGN_URL,
-					connectInstallation,
-				);
-			}
+			if (!isString(storedUrl)) return;
+			if (!this.areUrlsOfSameDesign(storedUrl, design.url)) return;
```
akostevich-atlassian, line 302: **note:** Compare URLs strictly by `fileKey` and `nodeId` components to fix the issue with the "Jira" widget in Figma being out of sync with Jira.

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
 	"compilerOptions": {
-		"target": "ES2020",
+		"target": "ES2021",
```
akostevich-atlassian, line 3: **note:** Change the target to allow using newer JS API (in particular `String.prototype.replaceAll`). Node 18, which we use, fully supports ES2021.

src/infrastructure/jira/jira-service.ts
```diff
@@ -222,10 +222,12 @@ class JiraService {
 		} catch (error) {
 			if (error instanceof NotFoundHttpClientError) {
```
akostevich-atlassian, line 223: **question:** @stannnous , what do you think about always overwriting `attached-design-url` on linking a design? Currently, the app writes to this property only if it is empty.

This could help to avoid data consistency issues in some edge cases (e.g., a design was unlinked but Connect app didn't remove the design URL from `attached-design-url`). I do not know such use cases but it looks like just a more reliable solution.

The side effect of this would be the "Jira" widget in Figma always showing the **last** linked design instead of the one linked from widget or the first one linked from Jira 🤯

stannnous: What does the existing Figma app do if you add a second design in Jira? Does it update the `attached-design-url` or keep the original value?

I think your suggestion makes sense but I would err on maintaining the existing behaviour.

stannnous (approved):

PR #171, opened 2023-11-20
FABB-2: Handle when users connect Figma teams with no paid plan

akostevich-atlassian:
The PR addresses comments in https://github.com/atlassian-labs/figma-for-jira/pull/153.

## Changes

- **refactor:** Handles paid plan-related `HTTP 400` errors from Figma API in `FigmaService.createFileUpdateWebhook` instead of `connectFigmaTeamUseCase`. See here for more detail: https://github.com/atlassian-labs/figma-for-jira/pull/153#discussion_r1398662659.
- **refactor:** Returns [HTTP 402 Payment Required](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/402) when a user connects Figma teams with no paid plan. This allows the frontend to easily distinguish plan-related errors from others.
- **feat:** Updates UI messages according to the latest requirements. 
- **test:** Adds additional unit and integration tests.

### DIFF
--- a/admin/jest.config.ts
+++ b/admin/jest.config.ts
@@ -5,9 +5,7 @@ const config: Config = {
 	testEnvironment: 'node',
 	maxWorkers: 4,
 	rootDir: '.',
-	roots: ['<rootDir>/src'],
-	testRegex: '(\\.|/)(test|spec)\\.tsx?$',
-	setupFiles: ['dotenv-expand/config'],
+	testRegex: '(\\.|/)test\\.tsx?$',
 	clearMocks: true,
 };
 
--- a/admin/src/pages/auth/auth-page.tsx
+++ b/admin/src/pages/auth/auth-page.tsx
@@ -69,7 +69,7 @@ export function AuthPage({ authorizationEndpoint }: AuthPageProps) {
 								flexDirection: 'column',
 							})}
 						>
-							<span>Admin permissions for a Figma team</span>
+							<span>Admin permissions for a paid Figma team</span>
 							<FigmaPermissionsPopup>
 								How to check Figma permissions
 							</FigmaPermissionsPopup>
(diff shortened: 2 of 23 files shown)


admin/jest.config.ts
```diff
@@ -0,0 +1,12 @@
+import type { Config } from 'jest';
```
akostevich-atlassian, line 1: **note:** Create a separate Jest config for the frontend project to avoid sharing the same configuration between frontend and backend (as they might be very different).

src/web/middleware/error-handler-middleware.ts
```diff
@@ -41,6 +42,11 @@ export const errorHandlerMiddleware = (
 			res.status(HttpStatusCode.Forbidden).send({ message: err.message });
 			return next();
 		}
+
+		if (err instanceof PaidFigmaPlanRequiredUseCaseResultError) {
+			res.status(HttpStatusCode.PaymentRequired).send({ message: err.message });
```
akostevich-atlassian, line 47: **note:** [HTTP 402 Payment Required](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/402) is a non-standard code but it is very convenient to communicate plan-related errors without introducing additional app-specific error codes.

tsconfig.json
```diff
@@ -19,5 +19,5 @@
 		"outDir": "build",
 		"baseUrl": "."
 	},
-	"include": ["src/**/*", "test/**/*"]
+	"include": ["src/**/*"]
```
akostevich-atlassian, line 22: **note:** Remove the accidentally found legacy artifact.

admin/src/pages/auth/auth-page.tsx
```diff
@@ -69,7 +69,7 @@ export function AuthPage({ authorizationEndpoint }: AuthPageProps) {
 								flexDirection: 'column',
 							})}
 						>
-							<span>Admin permissions for a Figma team</span>
+							<span>Admin permissions for a paid Figma team</span>
```
akostevich-atlassian, line 72: **note:** Updated the message according to the latest requirements from Jade.

stannnous (approved):

PR #173, opened 2023-11-21
NONE: Fix the issue with managing `attached-design-url-v2` Issue Property

akostevich-atlassian:
## Changes

- **fix:** Includes a design name into `url` when write design data to `attached-design-url-v2` for consistency with `attached-design-url` and compatibility with the "Jira" widget.

### DIFF
--- a/src/infrastructure/jira/jira-service.ts
+++ b/src/infrastructure/jira/jira-service.ts
@@ -209,9 +209,11 @@ export class JiraService {
 	 * @internal
 	 * Only visible for testing. Please use {@link saveDesignUrlInIssueProperties}
 	 */
-	// TODO: Consider removing this method since:
+	// TODO: Consider removing this method after deprecating the previous version of
+	//  "Figma for Jira" (which is included in this app as the fallback).
+	//  It should be safe to do so since:
 	//  - The "Jira" widget reads from `attached-design-url-v2`
-	//  - The previous version of the "Figma for Jira" does not set `attached-design-url`
+	//  - The previous version of the "Figma for Jira" did not set `attached-design-url` anyway
 	setAttachedDesignUrlInIssuePropertiesIfMissing = async (
 		issueIdOrKey: string,
 		design: AtlassianDesign,
--- a/src/infrastructure/jira/jira-service.ts
+++ b/src/infrastructure/jira/jira-service.ts
@@ -16,7 +16,10 @@ import {
 	SchemaValidationError,
 } from '../../common/schema-validation';
 import { ensureString, isString } from '../../common/string-utils';
-import { appendToPathname } from '../../common/url-utils';
+import {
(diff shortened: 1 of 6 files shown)


src/infrastructure/jira/jira-service.ts
```diff
@@ -247,31 +245,38 @@ class JiraService {
...
+		}
 
-		const newValue = [
-			...storedValueExcludingItemWithTargetUrl,
-			{ url, name: displayName },
-		];
+		let newValue = storedValue ? [...storedValue] : [];
+
+		if (storedItem) {
+			newValue = newValue.map((item) => (item === storedItem ? newItem : item));
```
stannnous, line 276: When the design already exists in the issue property, you decided to update it with the new value? Just confirming this is correct because we were discussing leaving the original URL unchanged and perhaps only updating the name.

stannnous (approved):

src/infrastructure/jira/jira-service.ts
akostevich-atlassian: Yes, sorry, I had to explain this on the PR.

This was my initial plan and implementation, but, unfortunately, it didn't fix the issue and I had to change the approach.

Please, consider the following scenario with the initially planned solution:

1. A user adds a design from the old experience
   ```js
   // Issue Properties
   {
      "attached-design-url-v2": [
        {
          url: "https://www.figma.com/file/abc/Cool-Design?mode=design",
          name: "Cool Design"
        }
      ]
   }
   ```
 2. A user switches to the new experience and backfills data:
   ```js
   // Issue Properties
   {
      "attached-design-url": "https://www.figma.com/file/abc", // Normalised
      "attached-design-url-v2": [
        {
          url: "https://www.figma.com/file/abc/Cool-Design?mode=design", // Original
          name: "Cool Design"
        }
      ]
   }
   ```  
 3. A user switches back to the old experience:
    ```js
   // Issue Properties
   {
      "attached-design-url": "https://www.figma.com/file/abc",
      "attached-design-url-v2": [
        {
          url: "https://www.figma.com/file/abc/Cool-Design?mode=design", // Original
          name: "Cool Design"
        },
        {
          url: "https://www.figma.com/file/abc", // Ooops! Added from `attached-design-url`.
          name: "Cool Design"
        }
      ]
   }
   ```

Therefore, while we write to `attached-design-url`, we have to write it **consistently** with `attached-design-url-v2`, so the same URL is written to both Issue Properties. This is what the PR addresses.

As we discussed, there is still an edge case when a single design can be duplicated (`attached-design-url`):

1. A user links a design using the "Jira" widget in Figma 
2. A switches from the new experience to the old one

There is a cure for this issue but it would make the logic complicated, so it might be better to avoid it. It is an edge case and, normally, should never affect our users (ideally, we should not switch back to the old experience).

PR #177, opened 2023-11-23
FABB-62 Replace design in issue properties with parent design when node not found

stannnous:
Addresses the issue with updating issue properties @akostevich-atlassian identified in the previous PR: https://github.com/atlassian-labs/figma-for-jira/pull/175#issuecomment-1822514542

### DIFF
--- a/src/infrastructure/jira/jira-service.ts
+++ b/src/infrastructure/jira/jira-service.ts
@@ -99,6 +99,7 @@ export class JiraService {
 
 	saveDesignUrlInIssueProperties = async (
 		issueIdOrKey: string,
+		figmaDesignIdToReplace: FigmaDesignIdentifier,
 		design: AtlassianDesign,
 		connectInstallation: ConnectInstallation,
 	): Promise<void> => {
@@ -110,6 +111,7 @@ export class JiraService {
 			),
 			this.updateAttachedDesignUrlV2IssueProperty(
 				issueIdOrKey,
+				figmaDesignIdToReplace,
 				design,
 				connectInstallation,
 			),
@@ -203,6 +205,7 @@ export class JiraService {
 	 */
 	updateAttachedDesignUrlV2IssueProperty = async (
 		issueIdOrKey: string,
+		figmaDesignIdToReplace: FigmaDesignIdentifier,
 		design: AtlassianDesign,
 		connectInstallation: ConnectInstallation,
 	): Promise<void> => {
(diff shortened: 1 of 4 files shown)


src/infrastructure/jira/jira-service.ts
```diff
@@ -215,7 +218,7 @@ export class JiraService {
 			);
 
 		const storedItem = storedValue?.find((item) =>
-			this.areUrlsOfSameDesign(item.url, design.url),
+			this.isUrlForDesign(item.url, figmaDesignIdToReplace),
```
akostevich-atlassian, line 221: **issue:** Sam, taking into account that the old experience allows to link the _same_ designs using different URLs, the Issue Property can contain more than 1 item to replace.

```
// Issue Properties
{
   "attached-design-url-v2": [
     {
       url: "https://www.figma.com/file/abc/Cool-Design?node-id=1-1",
       name: "Cool Design"
     },
     {
       url: "https://www.figma.com/file/abc/Cool-Design?node-id=1-1&mode=design", // Original
       name: "Cool Design"
     },
     {
       url: "https://www.figma.com/file/abc/Cool-Design?node-id=1-1&mode=dev",
       name: "Cool Design"
     }
   ]
}
```

Interestingly, our current implementation does not handle "duplicates" -- it always updates the first found item. Likely, it is just an omission that does not cause any issues. However, with the recent changes, we need to handle duplicates: otherwise, the backfill card will be shown until all duplicates are not overridden.

As a possible solution, what do you think about one of the following?

- Update `updateAttachedDesignUrlV2IssueProperty` to update all items that correspond to `figmaDesignIdToReplace`.  
- Update `updateAttachedDesignUrlV2IssueProperty` to deduplicate items before writing to the Issue Property (can be risky but very simple -- of the top of my head, I don't see why it can cause problems).

akostevich-atlassian (approved):

src/infrastructure/jira/jira-service.ts
stannnous: I tested with 3 designs in issue properties:

- https://www.figma.com/file/uZUWdb7Q0zBZmtMYfENNPY/Docs-in-Jira---Worksheet
- https://www.figma.com/file/uZUWdb7Q0zBZmtMYfENNPY/Docs-in-Jira---Worksheet?type=whiteboard&node-id=18-3546&t=yHv635RRQoJtotCY-0
- https://www.figma.com/file/uZUWdb7Q0zBZmtMYfENNPY/Docs-in-Jira---Worksheet?type=whiteboard&node-id=18-3546&t=yHv635RRQoJtotCY-0&mode=dev

where node-id 18-3546 is a deleted node.

and found that de-duping as you suggested solved a couple of problems:
- The frontend is also de-duping and only sends one backfill request for each extracted designId
- The new experience didn't show duplicate designs but allowing duplicates in the issue property meant the old experience does.

So I made the change to de-dupe issue properties. With this change, the above test results in a single design being ingested and a single entry in the issue property ensuring both old and new experiences are consistent.

PR #178, opened 2023-11-28
FABB-69: Return `HTTP 404` when associate a non-existent Figma File or Node

akostevich-atlassian:
## Changes

- **feat:** Updates the `POST entities/associateEntity` endpoint to return `HTTP 404 Not Found` when a Figma File or Node does not exist.
- **refactor:** Applies minor improvements to `FigmaService`.

### DIFF
--- a/src/infrastructure/figma/figma-service.ts
+++ b/src/infrastructure/figma/figma-service.ts
@@ -51,7 +51,6 @@ export class FigmaService {
 	/**
 	 * Returns Atlassian design for the Figma design with the given ID if it exists; otherwise -- `null`.
 	 *
-	 * @throws {DesignNodeNotFoundError} Design node referenced by the designId was not found.
 	 * @throws {UnauthorizedFigmaServiceError} Not authorized to access Figma.
 	 */
 	getDesign = async (
@@ -64,9 +63,9 @@ export class FigmaService {
 			const { fileKey, nodeId } = designId;
 
 			if (!nodeId) {
-				return this.getDesignForFile(fileKey, credentials);
+				return await this.getDesignForFile(fileKey, credentials);
 			} else {
-				return this.getDesignForNode(fileKey, nodeId, credentials);
+				return await this.getDesignForNode(fileKey, nodeId, credentials);
 			}
 		});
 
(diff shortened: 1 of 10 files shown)


src/usecases/associate-design-use-case.ts
```diff
@@ -46,6 +49,8 @@ export const associateDesignUseCase = {
 				jiraService.getIssue(associateWith.id, connectInstallation),
 			]);
 
+			if (!design) throw new FigmaDesignNotFoundUseCaseResultError();
```
akostevich-atlassian, line 52: **note:** The key change :)

src/infrastructure/figma/figma-service.ts
```diff
@@ -51,15 +49,15 @@ export class FigmaService {
 	};
 
 	/**
-	 * Return Atlassian design for the Figma design with the given ID.
+	 * Returns Atlassian design for the Figma design with the given ID if it exists; otherwise -- `null`.
 	 *
 	 * @throws {DesignNodeNotFoundError} Design node referenced by the designId was not found.
```
elainefigma, line 54: minor: Does this method still throw a DesignNodeNotFoundError? If no, should we delete this comment?

src/infrastructure/figma/figma-service.ts
```diff
@@ -344,50 +342,91 @@ export class FigmaService {
 		});
 
 	/**
-	 * Return Atlassian design representation for the given Figma File.
-	 *
-	 * @throws {UnauthorizedFigmaServiceError} Not authorized to access Figma.
+	 * Returns Atlassian design representation for the given Figma File if it exists; otherwise -- `null`.
```
elainefigma, line 345: minor: Can this method still throw an UnauthorizedFigmaServiceError i.e. if the user is not logged in to Figma? I have a similar question for the getDesignForNode method as well.

elainefigma (approved):
Thank you for fixing this! This will also help clean up our error monitors for these endpoints.

src/infrastructure/figma/figma-service.ts
akostevich-atlassian: Oh, great catch, @elainefigma . Fixed.

src/infrastructure/figma/figma-service.ts
akostevich-atlassian: These private methods currently do not do error translation but now I am thinking that they should :) I've updated these methods and the code comments accordingly. Thank you, @elainefigma , for the careful review!

PR #180, opened 2023-11-29
NONE: Resolve the issue with File-based design sync

akostevich-atlassian:
## Changes

- **fix:** Resolves the data sync issue for the scenario when a Figma File is linked but no Node from this File is linked within a Jira instance.


elainefigma (approved):
This looks good to me, thanks!

PR #181, opened 2023-11-30
FABB-69 Remove Figma dependency from disassociate design use case

stannnous:
Uses a stub design with updateSequenceNumber = 0 to remove the issue association in the disassociate design use case. This removes the dependency on fetching the Figma design for the disassociate use case which fixes the bug when the design has been deleted.

In the future we would like to provide an API to allow removing the issue association without sending any design data which will allow removal of the design stub.

### DIFF
--- a/src/usecases/disassociate-design-use-case.ts
+++ b/src/usecases/disassociate-design-use-case.ts
@@ -1,5 +1,4 @@
 import {
-	FigmaDesignNotFoundUseCaseResultError,
 	ForbiddenByFigmaUseCaseResultError,
 	InvalidInputUseCaseResultError,
 } from './errors';
@@ -8,6 +7,8 @@ import type { AtlassianEntity } from './types';
 import type { AtlassianDesign, ConnectInstallation } from '../domain/entities';
 import {
 	AtlassianAssociation,
+	AtlassianDesignStatus,
+	AtlassianDesignType,
 	buildJiraIssueUrl,
 	FigmaDesignIdentifier,
 	JIRA_ISSUE_ATI,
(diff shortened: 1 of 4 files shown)


src/web/routes/entities/integration.test.ts
```diff
@@ -959,75 +950,5 @@ describe('/entities', () => {
 				.set('Content-Type', 'application/json')
 				.expect(HttpStatusCode.BadRequest);
 		});
-
```
akostevich-atlassian, line 962: ☺️

src/usecases/disassociate-design-use-case.ts
```diff
@@ -49,16 +55,24 @@ export const disassociateDesignUseCase = {
...
-			const [design, issue] = await Promise.all([
-				figmaService.getDesign(figmaDesignId, {
-					atlassianUserId,
-					connectInstallationId: connectInstallation.id,
-				}),
-				jiraService.getIssue(disassociateFrom.id, connectInstallation),
-			]);
+			// By setting updateSequenceNumber to 0 the design will not be updated,
+			// and therefore we don't need to fetch the actual design data.
+			const designStub = {
```
akostevich-atlassian, line 60: **suggestion (non-blocking):** We could expose the desired API from `JiraService` and  encapsulate the implementation there.

For example:

```ts
export const disassociateDesignUseCase = {
   // ...
   
   // Under the hood, `JiraService.deleteDesignAssociation` creates the stub and use
   // available Jira API to implement the desired behavior. Later, this implementation can be
   // replaced with no effect on the use case.
   await jiraService.deleteDesignAssociation(entity.id, designIssueAssociation);
   
    // ...
}
```

PR #183, opened 2023-12-01
NONE: Use `URL` instead of `string` for URL-related function contract

akostevich-atlassian:
The PR contains refactoring extracted from FABB-69-related changes for convenience. The PR for FABB-69 changes will be created shortly.

## Changes

- **refactor:** Uses `URL` instead of `string` for a parameter and return type of URL-related functions. The usage of a more specifc value object helps to avoid repetitive parsing of the same URL among different functions and simplify error handling.


src/domain/entities/testing/mocks.ts
```diff
@@ -50,7 +50,7 @@ export const generateFigmaDesignUrl = ({
 	fileKey?: string;
 	nodeId?: string;
 	mode?: string;
-} = {}) => {
+} = {}): string => {
```
akostevich-atlassian, line 53: **note:** Just add a missing returned type.

src/domain/entities/figma-design-identifier.ts
```diff
@@ -33,18 +33,17 @@ export class FigmaDesignIdentifier {
 	 *
 	 * The method parses the URL and extracts data required for building {@link FigmaDesignIdentifier}.
 	 */
-	static fromFigmaDesignUrl = (url: string): FigmaDesignIdentifier => {
-		const parsedUrl = new URL(url);
-
-		const pathComponents = parsedUrl.pathname.split('/');
+	static fromFigmaDesignUrl = (url: URL): FigmaDesignIdentifier => {
```
akostevich-atlassian, line 36: **note:** Changes the contract to accept `URL` so the function should not worry about invalid URLs and related error handling.

src/infrastructure/figma/transformers/utils.ts
```diff
@@ -10,15 +10,15 @@ export const buildDesignUrl = ({
 }: {
 	fileKey: string;
 	nodeId?: string;
-}): string => {
+}): URL => {
```
akostevich-atlassian, line 13: **note:** Return `URL` as a higher-level and more specific type.

stannnous (approved):

PR #186, opened 2023-12-05
NONE: Fix issue with inconsistent unit test behaviour between CI and a local machine

akostevich-atlassian:
@stannnous identified an inconsistent unit test behavior between CI and a local machine. While all unit tests pass locally, unit tests that depend on `getConfig` (directly or indirectly) and do not mock it fail in CI.

The issue happened since some modules read `.env`, which is not available in CI but normally available locally. This behavior is caused by the generated `PrismaClient`, which implicitly read `.env` _on the module initialization_. 

image

See for more detail here:

- https://github.com/prisma/prisma/issues/10104#issuecomment-1310088721
- [Prisma Docs - How does Prisma use environment variables?](https://www.prisma.io/docs/guides/development-environment/environment-variables#how-does-prisma-use-environment-variables)

Unfortunately, there is no way to disable this unexpected side effect. However, there is a way to make Prisma use the specified `.env` file. Therefore, the PR updates the setup to consistently use `.env.test`, which is included in the repository and available on all the environments.

## Changes

- **test:** Updates the setup to consistently use `.env.test` for all types of tests regardless of the environment.
- **test:** Deletes the code for mocking `getConfig` in unit tests as unnecessary.


package.json
```diff
@@ -30,7 +30,7 @@
 		"test": "run-s test:unit test:it admin:test",
 		"test:it": "(dotenv -e .env.test -- prisma migrate reset -f) && (DOTENV_CONFIG_PATH=.env.test jest -c jest.config.integration.ts --runInBand)",
 		"test:it:ci": "run-s start:sandbox:test test:it stop:sandbox:test",
-		"test:unit": "jest -c jest.config.unit.ts"
+		"test:unit": "DOTENV_CONFIG_PATH=.env.test jest -c jest.config.unit.ts"
```
akostevich-atlassian, line 33: **note:** This is the actual fix. All the remaining changes are either related to clean-up or minor setup improvements.

stannnous (approved):
Great find. Thank you!

PR #191, opened 2023-12-06
[ci] Remove side effect of integration tests

elainefigma:
We run integration tests in the CI pipeline before we build the docker container - [see the Github workflow](https://github.com/atlassian-labs/figma-for-jira/blob/0e831bb5de24d88fb31388cdfa3a090ba94777f8/.github/workflows/actions.yml#L30). The integration tests calls `scripts/start-sandbox.sh`. 

I notice that `prisma migrate dev` has a side-effect. Per [the Prisma docs](https://www.prisma.io/docs/concepts/components/prisma-migrate/migrate-development-production#:~:text=If%20it%20detects%20changes%20to,new%20migration%20from%20these%20changes): If `prisma migrate dev` detects a change to the Prisma schema, it generates a new migration file from these changes.

This resulted in some spooky behavior on our end:
- Our database schema is a bit different from the one in schema.prisma since it's managed by our database team.
- Running the integration tests generated a migration file, which said to drop some important tables.
- The docker build step copied the migration file to our staging / prod containers.

image

This PR updates the integration test script to use `prisma migrate deploy` instead. That way, we won't automatically generate migration files as part of the build process. This reduces the risk of copying over auto-generated files in our staging / prod deploy process.

Note that we also use scripts/start-sandbox.sh for local development. This means there may be an extra step in local development if you make schema changes.

### DIFF
--- a/scripts/start-sandbox.sh
+++ b/scripts/start-sandbox.sh
@@ -22,4 +22,4 @@ until [ "$(docker inspect -f '{{.State.Health.Status}}' "$DB_CONTAINER_NAME")" =
 done
 
 echo 'Running DB migrations...'
-dotenv -e "$ENV_FILE" -- prisma migrate dev
+dotenv -e "$ENV_FILE" -- prisma migrate deploy


elainefigma:
Yes! To add some information:
- We use a temporary database for integration tests. It gets shut down afterwards. One small difference here is we have a slightly separate process for spinning up the database. The Figma repo CI pipeline cannot use the ./start-sandbox script since our CI pipeline doesn't support docker-compose. To get around this, we spin up the temporary database via another method. Our sandbox script looks like this
```
set -eu

ENV_FILE=.env.test

# Check if Postgres is running on port 5433.
until nc -zv localhost 5433; do
  sleep 1
done

echo 'Running DB migrations...'
dotenv -e "$ENV_FILE" -- prisma migrate deploy
```
- Correct! The 0_init migration is specific to our repository. Our database engineer had to create some tables when setting up the figma-for-jira DB. The 0_init migration describes the tables we created. That way, the prisma schema changes and our DB are up-to-date.

akostevich-atlassian:
@elainefigma , thank you very much for the explanation! I have reproduced the issue locally as well. Regardless of this, the PR definitely makes the setup safer due to removing unexpected side effects -- I'm all for this!

akostevich-atlassian (approved):

PR #193, opened 2023-12-12
FABB-72: Fix error handling for invalid file URLs

elainefigma:
Suppose you have a project URL like https://www.figma.com/files/project/176167247/Team-project?fuid=1166427116484924636

Suppose you go to a Jira issue > Link a design URL. Currently, this causes a 500 error in POST /entities/associateEntity.

This PR updates the error handling so we correctly log it as a 400 error.

Ticket: https://figlassian.atlassian.net/browse/FABB-72

## Testing
- I added unit tests
- I tested locally

before | after
--- | ---
image | image

### DIFF
--- a/src/domain/entities/figma-design-identifier.ts
+++ b/src/domain/entities/figma-design-identifier.ts
@@ -1,5 +1,3 @@
-import { InvalidInputUseCaseResultError } from '../../usecases/errors';
-
 const FIGMA_DOCUMENT_NODE_ID = '0:0';
 
 /**
@@ -45,9 +43,7 @@ export class FigmaDesignIdentifier {
 		const nodeId = url.searchParams.get('node-id')?.replace('-', ':');
 
 		if (!fileKey)
-			throw new InvalidInputUseCaseResultError(
-				`Received invalid Figma URL: ${url.toString()}`,
-			);
+			throw new Error(`Received invalid Figma URL: ${url.toString()}`);
 
 		return new FigmaDesignIdentifier(fileKey, nodeId);
 	};
--- a/src/domain/entities/figma-design-identifier.ts
+++ b/src/domain/entities/figma-design-identifier.ts
@@ -1,3 +1,5 @@
+import { InvalidInputUseCaseResultError } from '../../usecases/errors';
+
 const FIGMA_DOCUMENT_NODE_ID = '0:0';
 
 /**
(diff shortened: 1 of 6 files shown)


src/domain/entities/figma-design-identifier.ts
```diff
@@ -43,7 +45,9 @@ export class FigmaDesignIdentifier {
 		const nodeId = url.searchParams.get('node-id')?.replace('-', ':');
 
 		if (!fileKey)
-			throw new Error(`Received invalid Figma URL: ${url.toString()}`);
+			throw new InvalidInputUseCaseResultError(
```
akostevich-atlassian, line 48: **suggestion:** @elainefigma , please, consider throwing `XxxUseCaseResultError` only from use cases.

[UseCaseResultError](https://github.com/atlassian-labs/figma-for-jira/blob/2af3c50acb70b98431f002fa4f2cad9462b2e3a7/src/usecases/errors.ts#L5) represents a user-facing error, which is mapped directly to the corresponding HTTP status code. Therefore, it is better to throw it from use cases since only a particular use case knows what this or that scenario means within its context.

For example:

- For use cases tha take a URL as an _input_ and use `FigmaDesignIdentifier.fromFigmaDesignUrl` to parse it, it is expected to throw `InvalidInputUseCaseResultError` (which corresponds to `HTTP 400`) in case of invalid input.
- For use cases that take an _internally stored_ URL and use `FigmaDesignIdentifier.fromFigmaDesignUrl` to parse it, it is expected to throw `Error` (which corresponds to `HTTP 500`) in case of invalid URL.

In addition to this, limiting the usage of [UseCaseResultError] to use cases will help to avoid circle dependencies between the `useCases` and `domain` layers: `useCases` should depend on `domain` but not vice versa.

Therefore, please, consider one of the following for `FigmaDesignIdentifier.fromFigmaDesignUrl`:

- **[Recommended]** Throw just an `Error` (as for example, the built-in `new URL('bla')` does). In this case you can just wrap a `FigmaDesignIdentifier.fromFigmaDesignUrl` call in a use case in `try` / `catch` and re-throw  `InvalidInputUseCaseResultError`.
- Add another method `FigmaDesignIdentifier.fromFigmaDesignUrlOrUndefined` and return `undefinded` as an indicator that a given URL cannot be parsed.

src/usecases/associate-design-use-case.ts
akostevich-atlassian, line 1: **issue:** Please, update the `backfillDesignUseCase` since it has a similar logic.

akostevich-atlassian (approved):

src/domain/entities/figma-design-identifier.ts
elainefigma: Thanks for the review and the context. That makes sense to me. I updated the code.

PR #194, opened 2023-12-13
FABB-73: Add error handling for Jira getIssue 404

elainefigma:
Suppose you do not have permissions to access a particular Jira issue. Then, the Jira getIssue request gives you a 404. 

Right now, we're mapping this to 500 errors in associateEntity. However, since this is an expected case, this PR tweaks the error handling to map these to 404 errors in the endpoint.

Ticket: https://figlassian.atlassian.net/browse/FABB-73

### DIFF
--- a/src/infrastructure/jira/jira-service.test.ts
+++ b/src/infrastructure/jira/jira-service.test.ts
@@ -13,9 +13,10 @@ import {
 import type { AttachedDesignUrlV2IssuePropertyValue } from './jira-service';
 import {
 	ConfigurationState,
+	IssueNotFoundJiraServiceError,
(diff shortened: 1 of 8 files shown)


akostevich-atlassian (commented):
@elainefigma , while the changes looks good, please, do not merge the PR until we investigate a scenario when an `HTTP 404` error occurs. Please, see more detail in the ticket.

akostevich-atlassian:
@elainefigma , we have investigated the root cause of this issue. I've added more detail on our finding in Slack and created another PR with the fix:

https://github.com/atlassian-labs/figma-for-jira/pull/218

PR #195, opened 2023-12-13
FABB-75: Add error handling for the isAdmin method

elainefigma:
See: https://figlassian.atlassian.net/browse/FABB-75

We see some number of 500 errors in the GET /admin endpoint. The root cause is a 403 error in the jira-service isAdmin check.

image

This PR adds error handling for the isAdmin check. That way, if there's a 403 error, we can map that to a 401 error in the top-level endpoint.

## Testing
- I added unit tests.

### DIFF
--- a/src/infrastructure/jira/jira-service.test.ts
+++ b/src/infrastructure/jira/jira-service.test.ts
@@ -13,9 +13,10 @@ import {
 import type { AttachedDesignUrlV2IssuePropertyValue } from './jira-service';
 import {
 	ConfigurationState,
+	ForbiddenByJiraServiceError,
 	issuePropertyKeys,
-	JiraService,
 	jiraService,
+	JiraService,
 	SubmitDesignJiraServiceError,
 } from './jira-service';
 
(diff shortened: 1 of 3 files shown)


src/web/middleware/jira/jira-admin-only-auth-middleware.ts
```diff
@@ -27,5 +30,14 @@ export const jiraAdminOnlyAuthMiddleware = (
 
 			next();
 		})
-		.catch(next);
+		.catch((err) => {
+			if (err instanceof ForbiddenByJiraServiceError) {
+				return next(
+					new UnauthorizedResponseStatusError(
```
elainefigma, line 36: For this bit, we're mapping an underlying 403 error to a 401 error. I don't think we need to expose the underlying 403 error, so this seems alright to me.

src/infrastructure/jira/jira-service.test.ts
```diff
@@ -957,7 +961,7 @@ describe('JiraService', () => {
 			expect(result).toBe(false);
 		});
 
-		it('should return false if user has ADMINISTER permission', async () => {
+		it('should return true if user has ADMINISTER permission', async () => {
```
akostevich-atlassian, line 964: **praise:** Nice catch!

akostevich-atlassian:
**note:** @elainefigma , while these changes look good, frankly speaking, I cannot come up with a good scenario when a user can get `HTTP 403` from this Jira endpoint. Please, proceed with merging these changes (they are nice to have on their own!) but do not close the ticket yet.

akostevich-atlassian (approved):

PR #197, opened 2023-12-13
NONE: remove packages.atlassian.com dependency

Roystbeef:
This got flagged by one of our devx engineers. Our CI was failing bc package.atlassian.com was having an outage. IMO this is unexpected so putting up a PR to update our package-lock.json to remove this dependency.

Originally I tried to regenerated our package-lock.json w the following commands:
```sh
rm -rf node_modules
rm package-lock.json
npm install -E --force
```
but this resulted in some CI failures. Despite `-E` corresponding to:
> -E, --save-exact: Saved dependencies will be configured with an exact version rather than using npm's default semver range operator.
it was still altering our package-lock.json in unexpected ways.

So I tried installing the affected packages individually pinned to a specific version and noticed that only the resolved field gets updated. So I just ran find/replace on `https://packages.atlassian.com/api/npm/npm-remote` / `https://registry.npmjs.org`
## Test Plan
- assert that `package.atlassian.com` is not in our package-lock.json
- run `rm -rf node_modules` in `figma-for-jira` and `figma-for-jira/admin`
- run `npm ci` in `figma-for-jira` and see that it succeeds
- assert that the app still starts / works as expected
- assert that CI passes


Roystbeef:
hmm, i did not expect CI to fail here lmao

elainefigma (approved):
This seems reasonable to me! Before merging, I'd suggest also checking with the Atlassian team here as well.

akostevich-atlassian (approved):

PR #198, opened 2023-12-14
[jira-service] Truncate long display names

elainefigma:
We are getting some 500 errors in the endpoint POST /entities/associateEntity. It looks like sometimes, the display name is too long, and we get a validation error. See stack trace.
image

I think in this case, it's better for us to handle the error gracefully vs. returning a 400 response code. 

The other option here would be to handle the truncation / displayName length validation on the Jira side in the submitDesigns method.

## Testing
- I added a unit test for this case.

### DIFF
--- a/src/infrastructure/jira/jira-service.ts
+++ b/src/infrastructure/jira/jira-service.ts
@@ -12,7 +12,7 @@ import {
 	parseJsonOfSchema,
 	SchemaValidationError,
 } from '../../common/schema-validation';
-import { ensureString, isString } from '../../common/string-utils';
+import { ensureString, isString, truncate } from '../../common/string-utils';
 import { appendToPathname } from '../../common/url-utils';
 import type {
 	AtlassianDesign,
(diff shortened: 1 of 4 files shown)


src/infrastructure/jira/jira-service.ts
```diff
@@ -75,11 +75,15 @@ export class JiraService {
...
 		const associationsLastUpdated = new Date();
+
+		// Jira does not allow display names longer than 255 characters.
+		const MAX_DISPLAY_NAME_LENGTH = 255;
 		const response = await jiraClient.submitDesigns(
 			{
 				designs: designs.map(
 					({ design, addAssociations, removeAssociations }) => ({
 						...design,
+						displayName: design.displayName.slice(0, MAX_DISPLAY_NAME_LENGTH),
```
akostevich-atlassian, line 86: **suggestion (non-blocking):** Please, consider adding adding `…` to the end of the string to indicate that it was truncated.

Here how we could do this:

1. Add `truncate` function to `common/stringUtils` (inspired by [this](https://javascript.info/task/truncate)):
  ```ts
  function truncate(str, maxLength) {
    if (str.length <= maxLength) return str;
    
    return `${str.slice(0, maxlength - 1)}…`;
  }
  ```
2. Call `truncate(design.displayName, MAX_DISPLAY_NAME_LENGTH)` here.

akostevich-atlassian (approved):

src/infrastructure/jira/jira-service.ts
elainefigma: Good suggestion! I added the helper function and some tests.

PR #199, opened 2023-12-28
Reset app configuration state on install/uninstall

atlasctz:
Delete the 'is-configured' app property key in jira cloud when the f4j app is uninstalled

This is to set the configured state to 'NOT_SET' when a user uninstalls so the state does not persist across installations

The complexity to add this to the /install flow is not worth it. If the app property deletion fails on uninstall I imagine this would be quite rare and could be dealt with in future if it does become a problem.

### DIFF
--- a/src/infrastructure/jira/jira-client/jira-client.test.ts
+++ b/src/infrastructure/jira/jira-client/jira-client.test.ts
@@ -259,6 +259,22 @@ describe('JiraClient', () => {
 		});
 	});
 
+	describe('deleteAppProperty', () => {
+		const propertyKey = 'property-key';
+		it('should delete app property', async () => {
+			jest.spyOn(axios, 'delete').mockResolvedValue({
+				status: HttpStatusCode.NoContent,
+			});
+
+			await jiraClient.deleteAppProperty(propertyKey, connectInstallation);
+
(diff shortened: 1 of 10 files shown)


akostevich-atlassian (approved):
**note:** For context, the app cannot call Jira API from the "[Installed](https://developer.atlassian.com/cloud/jira/platform/connect-app-descriptor/#lifecycle)" lifecycle event handler since the installation is not finished at this point. As a result, all request to Jira API will result in `HTTP 400`. Please, see for here for more detail: https://community.developer.atlassian.com/t/jira-api-returns-400-all-the-time-in-the-installed-lifecycle-callback/58794/3

As a possible solution, we could use the "[Enabled](https://developer.atlassian.com/cloud/jira/platform/connect-app-descriptor/#lifecycle)" lifecycle event to do the configuration state clean-up. However, the additional complexity might not worth the price.

Currently, in the worst case, users will see the configuration state from the previous installation if the "Uninstalled" lifecycle event handling fails on the app property deletion. I believe this is an edge case, and it does not break the app but affects only the convenience of the configuration functionality. Therefore, the implementation in this PR is good enough.

**CC:** @stannnous

PR #200, opened 2024-01-02
Fix 500 error in dev_resources due to Url already exists

elainefigma:
In the past month, we got 865 errors with response code 500 in the endpoint POST /entities/associateEntity. ~200 of those errors are due to an error in the Figma POST /dev_resources endpoint.

If the dev_resource URL already exists, then the /dev_resources endpoint returns an error message: 'Url already exists' and a 400 status code. This PR updates the error handling so we treat these errors as a 400.

image

### DIFF
--- a/src/infrastructure/figma/figma-service.ts
+++ b/src/infrastructure/figma/figma-service.ts
@@ -201,6 +201,11 @@ export class FigmaService {
 					{ errors: response.errors },
 					'Failed to create Figma dev resources.',
 				);
+
+				const { error } = response.errors[0];
+				if (error === 'Url already exists') {
+					throw new InvalidInputFigmaServiceError(error);
+				}
 			}
 		});
 
--- a/src/infrastructure/figma/figma-service.test.ts
+++ b/src/infrastructure/figma/figma-service.test.ts
@@ -603,6 +603,33 @@ describe('FigmaService', () => {
 				}),
 			).rejects.toThrow(expectedError);
 		});
+
+		it('should throw when the dev_resource url already exists', async () => {
+			const issueKey = generateJiraIssueKey();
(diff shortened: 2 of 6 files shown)


akostevich-atlassian:
**note:** @elainefigma , the PR description says "dev_resources endpoint returns ... a 400 status code"

As far as I remember, `POST /v1/dev_resources` returns `HTTP 200` with `errors` in the body but not `HTTP 400`. The documentation says the following:

> If there are any dev resources that cannot be created, you may still get a `200` response. These resources will show up in the errors array. Some reasons a dev resource cannot be created include:
> - Resource points to a file_key that cannot be found.
> - The node already has the maximum of 10 dev resources.
> - Another dev resource for the **node has the same url**.

src/infrastructure/figma/figma-service.ts
```diff
@@ -201,6 +201,11 @@ export class FigmaService {
 					{ errors: response.errors },
 					'Failed to create Figma dev resources.',
 				);
+
+				const { error } = response.errors[0];
```
akostevich-atlassian, line 205: **question:** If I am not missing anything, based on the existing code, it seems `tryCreateDevResourceForJiraIssue` does not throw when `response.errors` is not empty. This being the case, could we check why does `POST /entities/associateEntity` return `HTTP 500` when dev resource creation fails with an error?

elainefigma:
@akostevich-atlassian Ah, you're right. If there is an error, POST /dev_resources does indeed return a 200 status code, so this is not the actual cause of the error. 

image

I'm going to close out this PR since it's not correct. I filed a separate Jira ticket for the root cause https://figlassian.atlassian.net/browse/FABB-79. In short, it looks like the JiraService submitDesigns method sometimes returns a 500 status code.

PR #201, opened 2024-01-05
[jira-client] Log additional context for connect installation

elainefigma:
We want to log some additional context information to debug an issue. See [Slack thread.](https://figma.slack.com/archives/C01PWBUQ1KQ/p1704434064481049?thread_ts=1704221284.999049&cid=C01PWBUQ1KQ)

This PR updates the function withAxiosErrorTranslation so we can log additional context information.

Relates to: https://figlassian.atlassian.net/browse/FABB-78

### DIFF
--- a/src/infrastructure/axios-utils.test.ts
+++ b/src/infrastructure/axios-utils.test.ts
@@ -1,6 +1,14 @@
+import type { AxiosResponse } from 'axios';
 import { AxiosError } from 'axios';
 
 import { withAxiosErrorTranslation } from './axios-utils';
+import {
+	BadRequestHttpClientError,
+	ForbiddenHttpClientError,
+	HttpClientError,
+	NotFoundHttpClientError,
+	UnauthorizedHttpClientError,
+} from './http-client-errors';
 
 describe('Axios utils', () => {
 	describe('withAxiosErrorTranslation', () => {
@@ -12,44 +20,31 @@ describe('Axios utils', () => {
 			await expect(result).rejects.toThrow('test error');
 		});
 
-		it('can throw a 400 bad request error', async () => {
-			const result = withAxiosErrorTranslation(() => {
-				throw new AxiosError('Bad request', '400');
-			});
+		it.each([
+			[BadRequestHttpClientError, 'Bad request', 400],
+			[UnauthorizedHttpClientError, 'Unauthorized', 401],
+			[ForbiddenHttpClientError, 'Forbidden', 403],
+			[NotFoundHttpClientError, 'Not found', 404],
+			[HttpClientError, 'Internal Server Error', 500],
+		])(
+			'should throw %s when Axios error with HTTP status %s is caught',
+			async (ErrorType, message, status) => {
+				const axiosError = new AxiosError(
+					message,
+					undefined,
+					undefined,
+					undefined,
+					{
+						status,
+					} as AxiosResponse,
+				);
 
-			await expect(result).rejects.toThrow('Bad request');
-		});
+				const result = withAxiosErrorTranslation(() => {
+					throw axiosError;
+				});
 
-		it('can throw a 401 unauthorized error', async () => {
-			const result = withAxiosErrorTranslation(() => {
-				throw new AxiosError('Unauthorized', '401');
-			});
-
-			await expect(result).rejects.toThrow('Unauthorized');
-		});
-
-		it('can throw a 403 forbidden error', async () => {
-			const result = withAxiosErrorTranslation(() => {
-				throw new AxiosError('Forbidden', '403');
-			});
-
-			await expect(result).rejects.toThrow('Forbidden');
-		});
-
-		it('can throw a 404 not found error', async () => {
-			const result = withAxiosErrorTranslation(() => {
-				throw new AxiosError('Not Found', '404');
-			});
-
-			await expect(result).rejects.toThrow('Not Found');
-		});
-
-		it('can throw a generic http client error', async () => {
-			const result = withAxiosErrorTranslation(() => {
-				throw new AxiosError('Internal Server Error', '500');
-			});
-
-			await expect(result).rejects.toThrow('Internal Server Error');
-		});
+				await expect(result).rejects.toBeInstanceOf(ErrorType);
+			},
+		);
 	});
 });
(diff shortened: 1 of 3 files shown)


src/infrastructure/axios-utils.ts
```diff
@@ -8,10 +8,14 @@ import {
...
 
-export const withAxiosErrorTranslation = async <T>(fn: () => Promise<T>) => {
+export const withAxiosErrorTranslation = async <T>(
+	fn: () => Promise<T>,
+	context?: unknown,
+) => {
 	try {
 		return await fn();
 	} catch (e: unknown) {
+		console.log('Error in withAxiosErrorTranslation', { context, error: e });
```
pkovalsky-atlassian, line 18: I'm not sure how Datadog client works but will this log `context` in the Datadog?

akostevich-atlassian: While it can work, it would be better to use `Logger` for consistency with other places.

Consider changing to the following:

```ts
getLogger().info(e, "An HTTP request to the external system failed.", context);
```

As an alternative option, you could add context to a thrown error itself, so it will be looged automatically by a global error handler: 

```ts
// axios-utils.ts
export const withAxiosErrorTranslation = async <T>(
	fn: () => Promise<T>,
	context?: unknown,
) => {
	try {
		return await fn();
	} catch (e: unknown) {
		throw translateAxiosError(e, context);
	}
};

const translateAxiosError = (error: unknown, context?: unknown): unknown => {
	if (!isAxiosError(error)) return error;

	switch (error.response?.status) {
		case HttpStatusCode.BadRequest:
			return new BadRequestHttpClientError(
				error.message,
				error.response.data,
				error,
				context,
			);
		case HttpStatusCode.Unauthorized:
			return new UnauthorizedHttpClientError(
				error.message,
				error.response.data,
				error,
				context,
			);
		case HttpStatusCode.Forbidden:
			return new ForbiddenHttpClientError(
				error.message,
				error.response.data,
				error,
				context,
			);
		case HttpStatusCode.NotFound:
			return new NotFoundHttpClientError(
				error.message,
				error.response.data,
				error,
				context,
			);
		default:
			return new HttpClientError(
				error.message,
				error.response?.data,
				error,
				context,
			);
	}
};
```

```ts
// http-client-errors.ts
import { CauseAwareError } from '../common/errors';

export class HttpClientError extends CauseAwareError {
	constructor(
		message?: string,
		readonly response?: unknown,
		readonly cause?: Error,
		context?: unknown,
	) {
		super(message, cause);
	}
}

// ...
```

elainefigma: Thanks for the review! Yes - console.log statements do log the fields to Datadog. 

Thanks for the code sample - previously, I was having some trouble getting the pino logs to show up. In order to get the context field, I added it like 
```
getLogger().info(
	{ err: e, context },
	'An HTTP request to the external system failed.',
);
```
image

src/infrastructure/axios-utils.test.ts
```diff
@@ -0,0 +1,55 @@
...
+		it('throws the error if it is not an axios error', async () => {
+			const result = withAxiosErrorTranslation(() => {
+				throw new Error('test error');
+			}, 'some context');
+
+			await expect(result).rejects.toThrow('test error');
+		});
+
+		it('can throw a 400 bad request error', async () => {
+			const result = withAxiosErrorTranslation(() => {
```
akostevich-atlassian, line 16: **issue:** @elainefigma , it seems that the test needs additional setup to test error handling for a specific HTTP code. If I understand correctly, currently, all the tests cover only [the default case](https://github.com/atlassian-labs/figma-for-jira/pull/201/files#diff-9b4d89988f79370e3599908d999bfe7e7202a3f22d1d399c42aa863235749915R56).

For example, in order to test `HTTP 400`, you could do the following:

```ts
		it('can throw a 400 bad request error', async () => {
			const result = withAxiosErrorTranslation(() => {
				throw new AxiosError('Bad request', undefined, undefined, undefined, {
					status: 400,
				} as AxiosResponse<unknown, any>);
			});

			await expect(result).rejects.toBeInstanceOf(BadRequestHttpClientError);
		});
```

Alternatively, you could combine similar tests into one and avoid repetitive test setup by using `it.each`:

```
it.each([
			[BadRequestHttpClientError, 'Bad request', 400],
			[UnauthorizedHttpClientError, 'Unauthorized', 401],
			[ForbiddenHttpClientError, 'Forbidden', 403],
			[NotFoundHttpClientError, 'Not found', 404],
		])(
			'should throw a %s when HTTP code is %s',
			async (ErrorType, message, status) => {
				const axiosError = new AxiosError(
					message,
					undefined,
					undefined,
					undefined,
					{
						status,
					} as AxiosResponse<unknown, any>,
				);

				const result = withAxiosErrorTranslation(() => {
					throw axiosError;
				});

				await expect(result).rejects.toBeInstanceOf(ErrorType);
			},
		);
```

akostevich-atlassian (approved):

src/infrastructure/jira/jira-client/jira-client.ts
```diff
@@ -72,8 +77,12 @@ class JiraClient {
 	getIssue = async (
 		issueIdOrKey: string,
 		connectInstallation: ConnectInstallation,
-	): Promise<GetIssueResponse> =>
-		withAxiosErrorTranslation(async () => {
+	): Promise<GetIssueResponse> => {
+		const context = {
+			baseUrl: connectInstallation.baseUrl,
+			clientKey: connectInstallation.clientKey,
```
pkovalsky-atlassian, line 83: During investigating a different issue we realised that it would really help if we also logged `issueIdOrKey`. Since you're already on it, could you please add `issueIdOrKey` to the context?

src/infrastructure/axios-utils.test.ts
elainefigma: Ah, excellent catch. Thanks for the code sample! I've updated the tests.

src/infrastructure/jira/jira-client/jira-client.ts
elainefigma: Yes - I added issueIdOrKey. 

I also added propertyKey as well in case it's helpful.

PR #203, opened 2024-01-17
RFC: Update timestamps to reflect dev status changes

Roystbeef:


src/usecases/handle-figma-file-update-event-use-case.ts
```diff
@@ -60,9 +60,37 @@ export const handleFigmaFileUpdateEventUseCase = {
 
 		if (!designs.length) return;
 
-		await jiraService.submitDesigns(
-			designs.map((design) => ({ design })),
-			connectInstallation,
-		);
+		// Update the timestamp for all the designs that updated their dev status
+		await Promise.all([
+			await jiraService.submitDesigns(
```
stannnous, line 65: @Roystbeef I think I'm missing something here. Can you provide some detail about how this process works?

From what I can tell:
* `figmaService.getAvailableDesignsFromSameFile` returns the node with the stored `devStatusLastModified` on line 47 above
* Then here you send the design to Jira and update the `devStatusLastModified` at the same time

Wouldn't this send a stale value to Jira?

src/infrastructure/figma/transformers/figma-node-transformer.ts
```diff
@@ -60,18 +64,39 @@ export const tryTransformNodeToAtlassianDesign = ({
...
+		// is not tracked at a node level
+		nodeLastModified = fileResponse.lastModified;
+	} else if (devStatusLastModified != null) {
+		// If the timestamp we have from changing the dev status is more recent than
+		// the node's last modified time, we should use that value instead.
+		if (
+			getUpdateSequenceNumberFrom(devStatusLastModified) >
+			getUpdateSequenceNumberFrom(nodeLastModified)
+		) {
+			nodeLastModified = devStatusLastModified;
```
stannnous, line 86: How can we be sure this won't cause a race condition where the `devStatusLastModified` and `fileResponse.lastModified` collide or are processed out of order?

src/usecases/handle-figma-file-update-event-use-case.ts
stannnous: Thank you for the description in the summary. I understand how this works now. My concern with this approach is it's a lot of complexity to workaround an issue that is better solved in Figmas API.

stannnous:
> If there's logic in Jira's ingestion to ignore any updates where from designs where the lastModified time is older than the one it already has, we can remove devStatusLastModified and simplify the logic here.

Jira handles out of order processing and will store entities with older lastModified (updateSequenceNumber in the Jira API) if it hasn't received that version before, or if it has received that version, Jira will ignore the update. While it should not cause a problem, it is unusual because you'd be sending updated data with an older version number.

src/usecases/handle-figma-file-update-event-use-case.ts
Roystbeef: > My concern with this approach is it's a lot of complexity to workaround an issue that is better solved in Figmas API.

Yeah I agree that having Figma's API handle this is most definitely the better long-term solution. My understanding of the situation is the current `lastModified` value we currently expose for nodes just surfaces a value that previously was only used privately for a different set of use cases. There's a few approaches we could take for this API going forward (eg. building it with the REST API in mind so that devstatus changes (among others) are reflected in the timestamp).

Unfortunately, IMO this is the only feasible solution to getting this ready by GA and IMO feels better scoped since any tech debt we're introducing is only on the relevant integration :/

Roystbeef:
> Jira handles out of order processing and will store entities with older lastModified (updateSequenceNumber in the Jira API) if it hasn't received that version before, or if it has received that version, Jira will ignore the update.

hmm, this sounds like we have quite a bit of leeway then. In theory, we wouldn't need to keep track of the `devStatusLastModified` value, but for ease of debugging/correctness, I think our service should try to honor the contract of only having `lastUpdated` go up (also I already wrote the code for it haha)

> While it should not cause a problem, it is unusual because you'd be sending updated data with an older version number.

There should be no difference in data when this occurs. Since neither the node nor its dev status will have been modified in this case.

src/usecases/handle-figma-file-update-event-use-case.ts
stannnous: Sounds good @Roystbeef and we're happy to defer to you as maintainer of the app however please consider this issue was not raised by customers but only found by us while testing. You could defer until Figmas API can be changed, keeping this workaround in your back pocket in case you receive feedback from customers.

PR #207, opened 2024-01-25
NONE: Update our webhook route to send a status code back immediately

Roystbeef:
We were noticing a lot of retries when handling webhook events. This is because Figma's webhook service has a 4 second timeout which the asynchronous code we were running in this route was blocking.

To address this, we are updating the route to respond with a 200 status code in the route after starting our asynchronous logic

Unfortunately, as part of this, we do need to add some specific testing logic in order for our integration tests to properly wait for all the asynchronous logic succeeds.

## Test Plan
- Assert that CI passes
- Assert that the webhook route responds a lot faster now


src/web/routes/figma/figma-router.ts
```diff
@@ -42,8 +43,25 @@ figmaRouter.post(
 
 				handleFigmaFileUpdateEventUseCase
```
akostevich-atlassian, line 44: **suggestion:** Consider wrapping the code for asynchronous execution with [setImmediate](https://nodejs.org/en/learn/asynchronous-work/understanding-setimmediate) to make the intention explicit and the code self-descriptive.

```ts
figmaRouter.post(
	'/webhook',
	requestSchemaValidationMiddleware(FIGMA_WEBHOOK_EVENT_REQUEST_SCHEMA),
	figmaWebhookAuthMiddleware,
	(req: FigmaWebhookEventRequest, res: FigmaWebhookEventResponse) => {
		const webhookId = req.body.webhook_id;
		const { figmaTeam } = res.locals;

		switch (req.body.event_type) {
			case 'FILE_UPDATE': {
				const { file_key } = req.body;

				setImmediate(async () => {
					try {
						await handleFigmaFileUpdateEventUseCase.execute(
							figmaTeam,
							file_key,
						);
						getLogger().info(
							{ webhookId },
							'Figma webhook callback succeeded.',
						);
					} catch (e) {
						getLogger().error(e, 'Figma webhook callback failed.', {
							webhookId,
						});
					}
 
        	// Since Jest does not have the ability to wait on any pending
				  // promises, we need to manage this ourselves to signal that the
				  // async code has finished running
				  if (process.env.NODE_ENV === 'test') {
           completePendingRouteExecutionForTests(); 
          }
				});

				// Immediately send a 200 back to Figma, before doing any of our own
				// async processing
				return res.sendStatus(HttpStatusCode.Ok);
			}
			default:
				return res.sendStatus(HttpStatusCode.Ok);
		}
	},
);
```

We could go even further and move the asynchronous code to `src/jobs/handle-figma-file-update-event.ts`. This will help to simplify a route handler and get the code ready for the long-term queue-based solution (as we will just need to call the job in the new code).  

```ts
figmaRouter.post(
	'/webhook',
	requestSchemaValidationMiddleware(FIGMA_WEBHOOK_EVENT_REQUEST_SCHEMA),
	figmaWebhookAuthMiddleware,
	(req: FigmaWebhookEventRequest, res: FigmaWebhookEventResponse) => {
		const webhookId = req.body.webhook_id;
		const { figmaTeam } = res.locals;

		switch (req.body.event_type) {
			case 'FILE_UPDATE': {
				const { file_key } = req.body;

				setImmediate(() => void handleFigmaFileUpdateEvent(req.body, figmaTeam));

				// Immediately send a 200 back to Figma, before doing any of our own
				// async processing
				return res.sendStatus(HttpStatusCode.Ok);
			}
			default:
				return res.sendStatus(HttpStatusCode.Ok);
		}
	},
);
```

src/web/routes/figma/figma-router.ts
```diff
@@ -6,6 +6,7 @@ import {
 	FIGMA_OAUTH2_CALLBACK_REQUEST_SCHEMA,
 	FIGMA_WEBHOOK_EVENT_REQUEST_SCHEMA,
 } from './schemas';
+import { completePendingRouteExecutionForTests } from './testing';
```
akostevich-atlassian, line 9: **suggestion:** Consider decoupling the application code from the test infrastructure. There can be several reasons for this:

- Testing infrastructure can contain side effects (e.g., overrides) and its own test-only dependencies. By importing  testing modules, we can implicitly bring these side effects to the app. It can be safer to avoid these dependencies.
- This dependency will cause certain test unnecessary modules to be included in the app bundle and Docker image.

Consider decoupling the code by using events (via the built-in [EventEmitter](https://nodejs.org/en/learn/asynchronous-work/the-nodejs-event-emitter)):

```ts
// Somewhere in the route handler.
eventBus.emit('figma.webhook.succeeded', { webhookId });
```
```ts
// Somewhere in tests.
function waitForEvent(eventName: string): Promise<void> {
	return new Promise((resolve, reject) => {
		eventBus.once(eventName, resolve);
	});
}

await waitForEvent('figma.webhook.succeeded');
```

Please, find the implementation of this approach as a convenient example here:
https://github.com/atlassian-labs/figma-for-jira/compare/poc/webhook-route-duration-events?expand=1

akostevich-atlassian (changes requested):
Please, see the comments.

src/web/routes/figma/figma-router.ts
Roystbeef: o this is very clever/clean!

akostevich-atlassian (approved):

PR #208, opened 2024-01-26
AMS-26036: Update static routes to perform jwt validation

Roystbeef:
We got a ticket from atlassian's security team RE: returning <400 status for https://figma-for-jira.figma.com/static/admin and https://figma-for-jira.figma.com/static/issue-panel/issue-panel.html when passed invalid jwts

This is a non-issue since we're not doing any server-side rendering on these routes and subsequent requests do perform JWT validation.

But it doesn't hurt to validate the token before serving these static html files

## Test Plan
- Assert that the legacy and admin config page still work
- Deploy on staging first and verify that everything still works before deploying to prod

### DIFF
--- a/src/web/middleware/jira/jira-query-symmetric-jwt-auth-middleware.ts
+++ b/src/web/middleware/jira/jira-query-symmetric-jwt-auth-middleware.ts
@@ -1,0 +1,38 @@
+import type { NextFunction, Request, RequestHandler, Response } from 'express';
+
+import { jiraContextSymmetricJwtTokenVerifier } from '../../../infrastructure/jira/inbound-auth';
+import { UnauthorizedResponseStatusError } from '../../errors';
+
+/**
+ * Authenticates requests using a symmetric JWT token received in a query param.
+ *
+ * In case of successful authentication, `connectInstallation` and `atlassianUserId` are set in locals.
+ *
+ * @remarks
+ * Context JWT tokens are sent in a query param when loading an iframe
+ *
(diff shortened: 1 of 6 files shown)


src/web/middleware/jira/jira-query-symmetric-jwt-auth-middleware.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * In case of successful authentication, `connectInstallation` and `atlassianUserId` are set in locals.
+ *
+ * @remarks
+ * Context JWT tokens are sent in a query param when loading an iframe
+ *
+ * @see https://developer.atlassian.com/cloud/jira/platform/understanding-jwt-for-connect-apps/#types-of-jwt-token
+ * @see https://community.developer.atlassian.com/t/action-required-atlassian-connect-vulnerability-allows-bypass-of-app-qsh-verification-via-context-jwts/47072
+ */
+export const jiraQuerySymmetricJwtAuthMiddleware: RequestHandler = (
```
akostevich-atlassian, line 17: **nitpick (non-blocking):** Consider renaming this to `jiraContextSymmetricJwtFromQueryAuthMiddleware` to highlight that it verifies a [context symmetric JWT](https://developer.atlassian.com/cloud/jira/platform/understanding-jwt-for-connect-apps/#types-of-jwt-token).

We could also name our middlewares based on their purpose  instead of the token type (e.g., `staticPageAuthMiddleware`) but we will need to rename other middlewares as well for consistency.

akostevich-atlassian (approved):
Great job, Roy! 🚀

PR #209, opened 2024-02-12
MDTZ-1378: Resolve the issue with backfill timeout

akostevich-atlassian:
## Changes

- **fix:** Resolves the issue with the "Backfill Design" use case failing due to a timeout, which happens when a user backfills a Node from a large Figma File.
  - Updates the "Backfill Design" use case to submit only a minimal design (constructed from URL) synchronously and then submit a full design asynchronously in a background job.
- **refactor:** Resolves minor naming inconsistencies.


src/jobs/submit-full-design.ts
stannnous, line 1: Suggestion: This could have it's own test. Probably worth it because you've removed the "full design" ingestion test from the use case test.

stannnous (approved):

src/jobs/submit-full-design.ts
akostevich-atlassian: Agree! 

✅ Done.

PR #211, opened 2024-02-12
MDTZ-1461 static resource auth

akostevich-atlassian:
## Changes

- **fix:** Resolves the issue with the "Designs" panel not being available when a user uses the **old** experience due to a `HTTP 401` error. Please, see the comments below for more detail.
  - Allows serving static content (e.g., HTML, CSS, etc.) with no authentication/authorization. 

image

## Notes

The issue is resolved by allowing serving static content (e.g., HTML, CSS, etc.) with no authentication/authorization. While the authentication has been recently introduced within https://github.com/atlassian-labs/figma-for-jira/pull/208, we might need to re-evaluate this decision for the following reasons:

- Adding authentication and authorization to static content routes increases latency (as it adds a request to the database and Jira API).
- Adding authentication and authorization to static content routes requires a careful route configuration (to exclude non-HTML files) and brings extra complexity.
- Static content does not have any user data and therefore, can be served publicly.
- Jira does not specifically handle `HTTP 401` from iframes -- it looks like a generic error, so returning `HTTP 401` does not bring much value to users.

Please, see the comments below for more detail.


src/web/middleware/jira/jira-iframe-symmetric-jwt-authentication-middleware.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * In case of successful authentication, `connectInstallation` and `atlassianUserId` are set in locals.
+ *
+ * @remarks
+ * Context JWT tokens are sent in a query param when loading an iframe
+ *
+ * @see https://developer.atlassian.com/cloud/jira/platform/understanding-jwt-for-connect-apps/#types-of-jwt-token
+ * @see https://community.developer.atlassian.com/t/action-required-atlassian-connect-vulnerability-allows-bypass-of-app-qsh-verification-via-context-jwts/47072
+ */
+export const jiraIframeSymmetricJwtAuthenticationMiddleware: RequestHandler = (
```
akostevich-atlassian, line 19: **note:** Represents a replacement for the deleted `jiraContextSymmetricJwtFromQueryAuthenticationMiddleware` (introduced within https://github.com/atlassian-labs/figma-for-jira/pull/208) as the correct way to authenticate requests from iframes.

The new middleware uses a different verifier (`jiraIframeOrServerToServerSymmetricJwtTokenVerifier` instead of `jiraContextSymmetricJwtFromQueryAuthenticationMiddleware`) and apples minor fixes to how a request is handled.

However, the middleware is **not** currently used and can be safely removed if the team agrees on serving static content without authentication. Please, see other notes for more detail.

src/web/routes/static/static-router.ts
```diff
@@ -2,24 +2,8 @@ import { Router, static as Static } from 'express';
...
-
 export const staticRouter = Router();
 
-staticRouter.use(
-	'/admin/index.html',
-	jiraContextSymmetricJwtFromQueryAuthenticationMiddleware,
-	jiraAdminOnlyAuthorizationMiddleware,
-);
-
-staticRouter.use(
```
akostevich-atlassian, line 18: **note:** Currently, all requests to `/static/issue-panel.html` fail with `HTTP 401`. The issue is caused by a bug in the original `jiraContextSymmetricJwtFromQueryAuthenticationMiddleware`.

image

src/web/routes/static/static-router.ts
```diff
@@ -2,24 +2,8 @@ import { Router, static as Static } from 'express';
...
 
-import {
-	jiraAdminOnlyAuthorizationMiddleware,
-	jiraContextSymmetricJwtFromQueryAuthenticationMiddleware,
-} from '../../middleware/jira';
-
 export const staticRouter = Router();
 
-staticRouter.use(
-	'/admin/index.html',
```
akostevich-atlassian, line 13: **note:** The auth issue is not reproducible with the admin page since this page is served from the `/static/admin` but not `static/admin/index.html` URL (see [here](https://github.com/atlassian-labs/figma-for-jira/blob/5489ea7b0d6088f75833a51c15cfa6e03b80dc99/src/atlassian-connect.ts#L109)) and therefore, the auth middlewares have no impact.

stannnous (approved):

src/web/routes/static/static-router.ts
Roystbeef: 🤦🏻‍♂️ the confirmation bias in me testing this code

Roystbeef (approved):

PR #214, opened 2024-02-20
Update truncate string util to handle utf-16 / utf-8 strings

elainefigma:
Some strings have a different length in utf-8 vs. utf-16 encoding. Javascript strings are utf-16, but Jira expects utf-8 strings. This PR fixes the truncate function to first encode JS strings as UTF8 before truncating.

I wrote a test that fails before the change and passes afterward. Test result against master for reference. 
image


jvrankul23 (approved):
image
Good old JavaScript 😆

PR #215, opened 2024-02-26
refactoring logging in jobs to use child logger for common properties

aimenhamed:


src/jobs/handle-figma-file-update-event.ts
```diff
@@ -10,21 +10,16 @@ export const handleFigmaFileUpdateEvent = async (
 	figmaTeam: FigmaTeam,
 ): Promise<void> => {
 	const { file_key: fileKey, webhook_id: webhookId } = requestBody;
+	const logger = getLogger().child({ job: JOB_NAME, webhookId });
```
akostevich-atlassian, line 13: **nitpick:** Since `JOB_NAME` is not used among multiple places anymore, consider simply inlining its value here and getting rid of the `JOB_NAME` constant.

```ts
const logger = getLogger().child({ job: 'handleFigmaFileUpdateEvent', webhookId });
```

The same is applicable to another changed file.

akostevich-atlassian (approved):
**praise:** Awesome job! This small improvement establishes a good pattern, which can be reused down the road 👍

akostevich-atlassian (approved):

PR #222, opened 2024-03-26
NONE: clean up logging

knguyen-figma:
The pino logger was logging the full request payload and authorization headers when Axios throws an error. This change uses the `redact` option of pino to redact those values in the logs. 

Unfortunately, the `fast-redact` library that `pino` uses does not support deep/nested wildcards, we have three levels of static wildcards for the Authorization headers.

### DIFF
--- a/src/infrastructure/logger.ts
+++ b/src/infrastructure/logger.ts
@@ -5,6 +5,14 @@ import { getConfig } from '../config';
 
 let logger: Logger | undefined;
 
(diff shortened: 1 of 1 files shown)


Roystbeef (approved):
LGTM!

akostevich-atlassian (commented):
Sorry for being late... Just wanted to say that it is a great job! 👍

PR #224, opened 2024-04-04
NONE: Polyfill node 20 string functions if necessary

Roystbeef:
Heh, we're still using node 18 for figma and this is managed globally for our work laptops. While we're using node 20 in production since it's getting built from the docker container. For the ease of development, this PR polyfills the node20 string functions we're using so that integration tests pass in development


Isaiah-Turner (approved):

src/common/string-utils.ts
```diff
@@ -1,3 +1,6 @@
+// eslint-disable-next-line import/no-unassigned-import
+import 'core-js/proposals/well-formed-unicode-strings';
```
akostevich-atlassian, line 2: **suggestion (non-blocking):** Consider adding a comment with the explanation, so we know that it is safe to remove it when it is time :)

```ts
// TODO: Remove the import and `core-js` dependency once the local dev environment is updated to Node.js 20+.
// eslint-disable-next-line import/no-unassigned-import
import 'core-js/proposals/well-formed-unicode-strings';
```

akostevich-atlassian (approved):

PR #228, opened 2024-04-17
MDTZ-1463 remove feature flag and old web panel assets

jvrankul23:
This feature flag is now fully rolled out and has not served `false` since mid-March.

This PR removes references to the flag in the app descriptor, and static assets for the old web panel.


akostevich-atlassian (approved):

akostevich-atlassian:
**praise:** Great job, Job!

PR #229, opened 2024-04-26
TGF-94 Add lastupdatedby and iconurl to atlassiandesign type

shadyrachid:
Currently the `figma-for-jira` integration submits designs to Atlassian when an associated file is updated or during a backfill operation. However, the submitted payload does not include information about the user that made the change to the file.
This PR introduces two fields to the payload that is submitted to Atlassian.
 - `lastUpdatedBy`: This field includes details about the user that made the change to a file. This is only populated when handling a `FILE_UPDATE` event. During other experiences (i.e. backfill, etc ...) the `lastUpdatedBy` field will remain `undefined` as that information is not available.
 - `iconUrl`: As the designs are ingested into a generic design schema, this field will have a url to the resource icon of the design submitted. As it stands, there is no resource specific icon for Figma, it will remain unset.
 
## Why are these changes made?
In order to include Figma designs in a user's list of Recent work items from 3rd parties. Adding these fields will help identify which user made the change.

## How are the two fields populated?
When handling a `FILE_UPDATE` event an additional api call to `/meta` is made which includes the `last_touched_by` field in the response. The `last_touched_by` field is mapped to `lastUpdatedBy`.


src/infrastructure/figma/figma-service.ts
```diff
@@ -140,6 +148,7 @@ export class FigmaService {
 					},
 					accessToken,
 				);
+				fileMetaResponse = await figmaClient.getFileMeta(fileKey, accessToken);
```
akostevich-atlassian, line 151: **suggestion:** Please, consider sending requests concurrently by using `Promise.all`:

```ts
	try {
		const nodeIds = designIds.map((id) => id.nodeId).filter(isString);
		const [fileResponse, fileMetaResponse] = await Promise.all([
			figmaClient.getFile(/* ... */),
			figmaClient.getFileMeta(fileKey, accessToken)
		]);

		return designIds
			.map((designId) => {
				if (!designId.nodeId) {
					return transformFileToAtlassianDesign({
						fileKey: designId.fileKey,
						fileResponse,
						fileMetaResponse,
					});
				} else {
					return tryTransformNodeToAtlassianDesign({
						fileKey: designId.fileKey,
						nodeId: designId.nodeId,
						fileResponse,
						fileMetaResponse
					});
				}
			})
			.filter(isNotNullOrUndefined);
	}
	catch (e) {
		if (e instanceof NotFoundHttpClientError) return [];
		throw e;
	}
```

src/infrastructure/figma/figma-service.ts
```diff
@@ -413,10 +424,16 @@ export class FigmaService {
 					credentials.accessToken,
 				);
 
+				const fileMetaResponse = await figmaClient.getFileMeta(
```
akostevich-atlassian, line 427: **suggestion:** Please, consider sending requests concurrently. See the comment above for more detail.

src/infrastructure/figma/figma-service.ts
```diff
@@ -446,14 +463,23 @@ export class FigmaService {
 					},
 					credentials.accessToken,
 				);
+				const fileMetaResponse = await figmaClient.getFileMeta(
```
akostevich-atlassian, line 466: **suggestion:** Please, consider sending requests concurrently. See the comment above for more detail.

src/infrastructure/figma/transformers/figma-file-transformer.ts
```diff
@@ -37,6 +40,8 @@ export const transformFileToAtlassianDesign = ({
 		status: AtlassianDesignStatus.NONE,
 		type: AtlassianDesignType.FILE,
 		lastUpdated: fileResponse.lastModified,
+		lastUpdatedBy: fileMetaResponse.file.last_touched_by,
+		iconUrl: getResourceIconUrl(),
```
akostevich-atlassian, line 44: **question:** @shadyrachid , please, consider ingesting `null` by default and fallback to [JiraDesignInfoProvider.logoUrl](https://github.com/atlassian-labs/figma-for-jira/blob/59eca2b05f19be69ecf21598bc9b8ad986a3eb71/src/atlassian-connect.ts#L119) or internal Atlassian icon **on read** within Atlassian systems for the following reasons:

- `getResourceIconUrl` returns a URL to the icon from app static resources. This URL is internal to the app and can be changed at any time due to refactoring/restructuring. The app should not implicitly commit to maintain this URL over time -- it commits only to providing an icon URL through `jiraDesignInfoProvider.logoUrl`, which can use a different URL at different points of time.
- Any stored URL can get invalid at some point. Falling back to using `jiraDesignInfoProvider.logoUrl` on read allows to show the **current relevant** provider icon.

src/infrastructure/figma/figma-backfill-service.ts
```diff
@@ -55,7 +55,7 @@ export class FigmaBackfillService {
 			type: designId.nodeId
 				? AtlassianDesignType.NODE
 				: AtlassianDesignType.FILE,
-			lastUpdated: new Date(0).toISOString(), // Cannot be derived from the URL. Use the default value.
+			lastUpdated: new Date(0).toISOString(), // Cannot be derived from the URL. Use the default value.,
```
akostevich-atlassian, line 58: **nitpick:** It seems like an unintentional change :)

src/domain/entities/atlassian-design.ts
```diff
@@ -42,5 +43,7 @@ export type AtlassianDesign = {
 	readonly status: AtlassianDesignStatus;
 	readonly type: AtlassianDesignType;
 	readonly lastUpdated: string;
+	readonly lastUpdatedBy?: FigmaUser;
```
akostevich-atlassian, line 46: **question:** From what I see in the `Design` type declaration in Atlassian's codebase, `lastUpdatedBy` is of the following type (converted to TS for convenience):

```ts
interface User {
  accountId: string|null;
}
```

Taking this into account, please, could you double-check that the data the app ingests matches the expected contract? From what I see:
- `FigmaUser` does not have `accountId`.
- `FigmaUser` has an extra `email` field with sensitive personal information. Even if sending an extra field does not break anything, it is better to avoid sending this information among services.

If there is mismatch, consider adding another type `AtlassianProviderUser` before `AtlassianDesign` (by analogue with `AtlassianDesignStatus`) as it is _specific_ to `AtlassianDesign`. Then, we could transform `FigmaUser` to `AtlassianProviderUser` in transformers (e.g., `transformFileToAtlassianDesign`).

Please, let me know if you have any concerns.

akostevich-atlassian (changes requested):
Please, see other comments.

src/domain/entities/atlassian-design.ts
shadyrachid: Good call, the aim is to map the external id in FigmaUser to accountId during ingestion.
Having `AtlassianProviderUser` and transforming it by stripping the email and only including the external id is the better approach. Will update it shortly.

src/infrastructure/figma/transformers/figma-file-transformer.ts
```diff
@@ -37,6 +39,7 @@ export const transformFileToAtlassianDesign = ({
 		status: AtlassianDesignStatus.NONE,
 		type: AtlassianDesignType.FILE,
 		lastUpdated: fileResponse.lastModified,
+		lastUpdatedBy: fileMetaResponse.file.last_touched_by,
```
akostevich-atlassian, line 42: **issue:** Please, do a manual mapping here as well to include only necessary field and exclude unnecessary (e.g., as `email`).

src/infrastructure/figma/transformers/figma-node-transformer.ts
```diff
@@ -72,6 +81,7 @@ export const tryTransformNodeToAtlassianDesign = ({
 			: AtlassianDesignStatus.NONE,
 		type: mapNodeTypeToDesignType(node.type),
 		lastUpdated: extra.lastModified,
+		lastUpdatedBy: fileMetaResponse.file.last_touched_by,
```
akostevich-atlassian, line 84: **issue:** Please, do a manual mapping here as well to include only necessary field and exclude unnecessary (e.g., as `email`).

src/infrastructure/figma/transformers/figma-file-meta-transformer.ts
```diff
@@ -19,6 +23,14 @@ type TransformFileMetaToAtlassianDesignParams = {
 	readonly fileMetaResponse: GetFileMetaResponse;
 };
 
+const transformFigmaUserToAtlassianProviderUser = (
```
akostevich-atlassian, line 26: **suggestion (non-blocking):** It is okay to map a single field manually within `transformFileMetaToAtlassianDesign` without calling a separate function (`transformFigmaUserToAtlassianProviderUser`).

Please, pay attention that if you would like to encapsulate the mapping in a function, it should be moved out of this module as the mapping is required in 3 places:

- `transformFileMetaToAtlassianDesign`
- `transformFileToAtlassianDesign`
- `transformNodeToAtlassianDesign`

src/domain/entities/atlassian-design.ts
```diff
@@ -15,6 +15,10 @@ export enum AtlassianDesignType {
 	OTHER = 'OTHER',
 }
 
+export interface AtlassianProviderUser {
+	readonly id: string;
```
akostevich-atlassian, line 19: **question:** Should it be `accountId`?

akostevich-atlassian: Got it! Thank you, @shadyrachid !

src/infrastructure/figma/transformers/figma-node-transformer.ts
shadyrachid: Sorry missed that 🤦

src/infrastructure/figma/transformers/figma-file-meta-transformer.ts
shadyrachid: Updated to use a dedicated user transformer. Tests also updated to check only id is returned.

akostevich-atlassian (approved):

src/infrastructure/figma/figma-client/testing/mocks.ts
```diff
@@ -113,10 +114,12 @@ export const generateGetMeResponse = ({
 export const generateGetFileMetaResponse = ({
 	name = generateFigmaFileName(),
 	lastModified = new Date(),
+	lastModifiedBy = generateFigmaUser(),
```
akostevich-atlassian, line 117: **nitpick:** Please, use another generation function (e.g., `generateLastTouchedBy`) or just assign a default value directly here (e.g., `{ id: uuidv4() }`).

src/infrastructure/figma/figma-client/schemas.ts
```diff
@@ -82,6 +82,14 @@ export const GET_FILE_META_RESPONSE_SCHEMA: JSONSchemaTypeWithId<GetFileMetaResp
 				properties: {
 					name: { type: 'string' },
 					last_touched_at: { type: 'string' },
+					last_touched_by: {
+						type: 'object',
+						properties: {
+							id: { type: 'string' },
+							email: { type: 'string' },
```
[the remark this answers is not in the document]
shadyrachid: Removed

src/infrastructure/figma/figma-service.test.ts
```diff
@@ -146,13 +147,18 @@ describe('FigmaService', () => {
 			const node = generateChildNode({ id: nodeId });
 			const designId = generateFigmaDesignIdentifier({ nodeId });
 			const credentials = generateFigmaOAuth2UserCredentials();
+			const figmaUser = generateFigmaUser();
```
akostevich-atlassian, line 150: **nitpick:** Please, see https://github.com/atlassian-labs/figma-for-jira/pull/229/files#r1585903139.

src/infrastructure/figma/transformers/figma-node-transformer.test.ts
```diff
@@ -72,10 +80,15 @@ describe('tryTransformNodeToAtlassianDesign', () => {
 			lastModified: irrelevantLastModified,
 		});
 
+		const fileMetaResponse = generateGetFileMetaResponse({
```
akostevich-atlassian, line 83: **nitpick:** Please, see https://github.com/atlassian-labs/figma-for-jira/pull/229/files#r1585903139.

We don't really need to generate `lastModifiedBy` value here as we are not interested in a specific value for this test -- any value is okay.

Therefore, consider using just this:

```ts
const fileMetaResponse = generateGetFileMetaResponse();

// ...
const result = tryTransformNodeToAtlassianDesign({
  fileKey,
  nodeId: node.id,
  fileResponse,
  fileMetaResponse,
});

// ...

expect(result).toStrictEqual({
  // ..
  lastUpdatedBy: { id: fileMetaResponse.lastModifiedBy.id },
});
```

akostevich-atlassian (approved):

src/infrastructure/figma/figma-service.ts
```diff
@@ -403,20 +416,24 @@ export class FigmaService {
...
+				]);
 
 				return tryTransformNodeToAtlassianDesign({
 					fileKey,
 					nodeId,
 					fileResponse,
+					fileMetaResponse,
 				});
 			} catch (e) {
 				if (e instanceof NotFoundHttpClientError) return null;
```
Roystbeef, line 439: non-blocking: is this still catching properly / returning null after the update to use `Promsise.all()`? idr if there's a specific unit/integration test to check this path

(ditto for any other `Promise.all()` changes in this PR)

Roystbeef (approved):
very cool!

src/infrastructure/figma/figma-service.ts
shadyrachid: There are tests for checking if it returns null for this code path.
This block still catches exceptions thrown  even after adding `Promise.all` and the type is not affected.

PR #231, opened 2024-05-07
NONE: Update last_touched_by schema validation to be nullable

Roystbeef:
We started seeing logs / 500 errors on `POST /associate/entities` because we were failing the schema validation for `GetFileMetaResponse`.

This is because `last_touched_by` can be `null` (eg. the user has since been deleted) which is different from the field being optional

This PR updates the app to be able to handle when `GetFileMetaResponse[file][last_touched_by]` is `null` and adds an integration test to check that this is handled properly

## Test Plan
- Revert all changes except for the integration test and assert that the integration test fails
- Apply the changes in this PR and assert that the integration test passes

### DIFF
--- a/src/infrastructure/figma/transformers/figma-file-meta-transformer.ts
+++ b/src/infrastructure/figma/transformers/figma-file-meta-transformer.ts
@@ -27,6 +27,7 @@ export const transformFileMetaToAtlassianDesign = ({
 	fileMetaResponse,
 }: TransformFileMetaToAtlassianDesignParams): AtlassianDesign => {
 	const designId = new FigmaDesignIdentifier(fileKey);
+	const lastUpdatedBy = fileMetaResponse.file.last_touched_by;
 
 	return {
 		id: designId.toAtlassianDesignId(),
@@ -37,7 +38,7 @@ export const transformFileMetaToAtlassianDesign = ({
 		status: AtlassianDesignStatus.NONE,
 		type: AtlassianDesignType.FILE,
 		lastUpdated: fileMetaResponse.file.last_touched_at,
-		lastUpdatedBy: { id: fileMetaResponse.file.last_touched_by.id },
+		...(lastUpdatedBy ? { lastUpdatedBy: { id: lastUpdatedBy.id } } : {}),
 		updateSequenceNumber: getUpdateSequenceNumberFrom(
 			fileMetaResponse.file.last_touched_at,
 		),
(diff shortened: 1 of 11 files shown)


knguyen-figma (approved):

src/infrastructure/figma/transformers/figma-file-meta-transformer.ts
```diff
@@ -37,7 +38,7 @@ export const transformFileMetaToAtlassianDesign = ({
 		status: AtlassianDesignStatus.NONE,
 		type: AtlassianDesignType.FILE,
 		lastUpdated: fileMetaResponse.file.last_touched_at,
-		lastUpdatedBy: { id: fileMetaResponse.file.last_touched_by.id },
+		...(lastUpdatedBy ? { lastUpdatedBy: { id: lastUpdatedBy.id } } : {}),
```
akostevich-atlassian, line 41: **suggestion:** Consider using a simpler expression:

```
return {
  // ...
  lastUpdatedBy: {
    id: fileMetaResponse.file.last_touched_by?.id ?? undefined
  },
  // ...
}
```

src/infrastructure/figma/transformers/figma-file-transformer.ts
```diff
@@ -39,7 +40,7 @@ export const transformFileToAtlassianDesign = ({
 		status: AtlassianDesignStatus.NONE,
 		type: AtlassianDesignType.FILE,
 		lastUpdated: fileResponse.lastModified,
-		lastUpdatedBy: { id: fileMetaResponse.file.last_touched_by.id },
+		...(lastUpdatedBy ? { lastUpdatedBy: { id: lastUpdatedBy.id } } : {}),
```
akostevich-atlassian, line 43: **suggestion:** Please, see https://github.com/atlassian-labs/figma-for-jira/pull/231/files#r1593215557.

src/infrastructure/figma/transformers/figma-node-transformer.ts
```diff
@@ -81,7 +82,7 @@ export const tryTransformNodeToAtlassianDesign = ({
 			: AtlassianDesignStatus.NONE,
 		type: mapNodeTypeToDesignType(node.type),
 		lastUpdated: extra.lastModified,
-		lastUpdatedBy: { id: fileMetaResponse.file.last_touched_by.id },
+		...(lastUpdatedBy ? { lastUpdatedBy: { id: lastUpdatedBy.id } } : {}),
```
akostevich-atlassian, line 85: **suggestion:** Please, see https://github.com/atlassian-labs/figma-for-jira/pull/231/files#r1593215557.

src/web/routes/entities/integration.test.ts
```diff
@@ -447,6 +447,136 @@ describe('/entities', () => {
 				},
 			]);
 		});
+		it('should handle when the last touched by is not available', async () => {
```
akostevich-atlassian, line 450: **suggestion (non-blocking):** This seems like use case, which can be cheaper and easier to cover with unit tests, as it's related to a special case and its handling is fully encapsulated in `transformers`. 

Please, consider, adding unit tests to `transformers` instead.

akostevich-atlassian:
**praise:** Great catch, @Roystbeef !

PR #233, opened 2024-05-28
MDTZ-1749: Improve naming and contract of various components

akostevich-atlassian:
The PR includes various changes extracted from another working branch (more detail soon) for reducing the scope of changes and code review convenience.

### Changes

- **feat:** Returns `HTTP 400` when a design is being associated not with a Jira Issue (as an unexpected scenario). 
- **refactor:** Updates `JiraDesignIssuePropertyService.deleteFromAttachedDesignUrlV2IssueProperties` to accept only design identifier instead of the whole `AtlassianDesign`.
- **refactor:** Improves naming for use case parameters to highlight that only Design-Issue association are handled. 
- **refactor:** Moves some constants/types closer to their usage.


src/usecases/associate-design-use-case.ts
```diff
@@ -20,7 +19,10 @@ import { associatedFigmaDesignRepository } from '../infrastructure/repositories'
 
 export type AssociateDesignUseCaseParams = {
 	readonly designUrl: URL;
-	readonly associateWith: AtlassianEntity;
+	readonly associateWithIssue: {
```
akostevich-atlassian, line 22: **note:** Make explicit that a design can be associated to a Jira Issue only.

src/usecases/disassociate-design-use-case.ts
```diff
@@ -23,11 +18,14 @@ import { jiraService } from '../infrastructure/jira';
 import { associatedFigmaDesignRepository } from '../infrastructure/repositories';
 
 export type DisassociateDesignUseCaseParams = {
-	readonly entity: {
+	readonly design: {
```
akostevich-atlassian, line 21: **note:** Make explicit that association is handled only for a Design and Jira Issue.

src/web/routes/entities/schemas.ts
```diff
@@ -22,7 +21,16 @@ export const ASSOCIATE_ENTITY_REQUEST_SCHEMA: JSONSchemaTypeWithId<{
 					},
 					required: ['url'],
 				},
-				associateWith: ATLASSIAN_ENTITY_SCHEMA,
+				associateWith: {
+					type: 'object',
+					properties: {
+						ati: { type: 'string', pattern: 'ati:cloud:jira:issue' }, // Handle only associations with a Jira Issue.
```
akostevich-atlassian, line 27: **note:** Return `HTTP 400 Bad Request` if `ati` is not `ati:cloud:jira:issue` on the entry point to the app. Therefore, the remaining code does not need to worry about the type of entity a Design is associated to.

src/domain/entities/atlassian-design.ts
```diff
@@ -22,7 +20,7 @@ export class AtlassianAssociation {
 	) {}
 
 	static createDesignIssueAssociation(issueAri: string): AtlassianAssociation {
-		return new AtlassianAssociation(JIRA_ISSUE_ATI, [issueAri]);
+		return new AtlassianAssociation('ati:cloud:jira:issue', [issueAri]);
```
akostevich-atlassian, line 23: **note:** Remove the constant and its usages to avoid this concept being leaked to other modules. The whole concept of `AtlassianAssociation` will be removed shortly due to the simplification on Atlassian side (work in progress).

Roystbeef (approved):
Thanks for cleaning this up! This all LGTM!

PR #236, opened 2024-05-30
NONE: Resolve an issue with unit test execution being dependent on previous tests

akostevich-atlassian:
### Changes

- **test:** Updates Jest configuration for stricter test isolation: automatically restore mock state before every test. See Jest [restoreMocks](https://archive.jestjs.io/docs/en/next/configuration#restoremocks-boolean) configuration property for more information.
- **test:** Fixes tests having an implicit dependency on other tests (which started to fail with the new stricter configuration).
- **test:** Updates tests not to use deprecated Jest API.
- **refactor:** Removes unnecessary `getLogger` mocking for unit tests.


jest.config.ts
```diff
@@ -8,7 +8,8 @@ const config: Config = {
 	roots: ['<rootDir>/src'],
 	testRegex: '(\\.|/)(test|spec)\\.tsx?$',
 	setupFiles: ['dotenv-expand/config'],
-	clearMocks: true,
+	restoreMocks: true,
```
akostevich-atlassian, line 11: `restoreMocks` is a stricter version of `clearMocks` and allows to catch implicit dependencies among tests.

According to the docs, [clearMocks](https://archive.jestjs.io/docs/en/next/configuration#clearmocks-boolean), in turn:

> Automatically clear mock calls and instances before every test. ... This does **not** remove any mock implementation that may have been provided.

And stricter [restoreMocks](https://archive.jestjs.io/docs/en/next/configuration#restoremocks-boolean):

> Automatically restore mock state before every test. ... This will lead to any mocks having their fake implementations **removed and restores** their initial implementation.

**CC:** @jvrankul23

Roystbeef (approved):
thank you for the cleanup here!!!

PR #242, opened 2024-08-06
NONE: Fix submit design response schema validation

Roystbeef:
This PR fixes the 500s we're seeing when associating designs in Jira. It looks like the response we're getting back changed and as a result of us failing the schema validation, we were throwing 500s in the route.

This PR updates the response schema type to the new values we're seeing


knguyen-figma (approved):

Roystbeef:
hmmm, not sure why CI is struggling w setting up the test db but I validated locally that `npm run test` is passing

PR #248, opened 2024-11-27
HL-1722 Update mutation APIs to receive userId in body

stannnous:
We are updating POST and PUT APIs to receive all input params in the request body instead of some as the query params. This involves moving the `userId` query param to the body for the following API endpoints:
* getEntityByUrl
* onEntityAssociated
* onEntityDisassociated

This is being done for compatibility reasons on the Atlassian platform which has already been updated to send these values in the body (it is currently sending both query and body params with the query to be removed at a later date).


akostevich-atlassian (approved):
Great job, Sam!

PR #249, opened 2025-01-10
NONE: Figma OAuth updates

ehasama-figma:
Figma is [encouraging our REST API clients](https://www.figma.com/developers/api#oauth_migration_guide) to use the Basic auth header to transmit OAuth credentials and use token exchange endpoints defined on `api.figma.com`. So this PR makes these updates.

## Test Plan
- updated integration tests such that they pass
- unit tests pass
- manually tested authenticating locally


Roystbeef (approved):
LGTM!

PR #254, opened 2025-05-05
NONE: env variable updates to support running on Figma's development and staging servers

knguyen-figma:
This pull request introduces support for a configurable base path and domain for the application, improving flexibility and compatibility with different deployment environments. The most significant changes include adding utility functions for generating app paths and URLs, updating static asset references across the frontend, and modifying backend routes and configurations to use the new base path.

The app should continue to work with the existing environment variables, however we will fallback to env variables that are prefixed with `FIGMA_FOR_JIRA_` if they are missing.

### DIFF
--- a/admin/src/api/axiosInstance.ts
+++ b/admin/src/api/axiosInstance.ts
@@ -1,8 +1,10 @@
 import axios from 'axios';
+import { getAppBasePath } from '../utils';
 
 const THIRTY_SECONDS_IN_MS = 30_000;
 const axiosRest = axios.create({
 	timeout: THIRTY_SECONDS_IN_MS,
+	baseURL: getAppBasePath() ?? undefined,
 });
 // Adding the token in the headers through interceptors because it is an async value
 axiosRest.interceptors.request.use(async (config) => {
--- a/admin/src/utils/parseTeamIdFromFigmaUrl.ts
+++ b/admin/src/utils/parseTeamIdFromFigmaUrl.ts
@@ -1,5 +1,9 @@
 const VALID_FIGMA_ORIGINS = ['www.figma.com', 'figma.com'];
 
+if (import.meta.env.VITE_FIGMA_FOR_JIRA_FIGMA_WEB_DOMAIN) {
+	VALID_FIGMA_ORIGINS.push(import.meta.env.VITE_FIGMA_FOR_JIRA_FIGMA_WEB_DOMAIN);
+}
+
 function withHttp(url: string) {
 	return !/^https?:\/\//i.test(url) ? `http://${url}` : url;
 }
--- a/src/atlassian-connect.ts
+++ b/src/atlassian-connect.ts
@@ -1,4 +1,4 @@
-import { getConfig } from './config';
+import { getConfig, getAppPath, getAppUrl, getFigmaDomain } from './config';
 
 const APP_NAME = 'Figma for JIRA Cloud';
 
@@ -64,8 +64,8 @@ export const connectAppDescriptor = {
 	 *
 	 */
 	lifecycle: {
-		installed: '/lifecycleEvents/installed',
-		uninstalled: '/lifecycleEvents/uninstalled',
+		installed: getAppPath('/lifecycleEvents/installed'),
+		uninstalled: getAppPath('/lifecycleEvents/uninstalled'),
 	},
 
 	/**
@@ -79,7 +79,7 @@ export const connectAppDescriptor = {
 			name: {
 				value: 'Configure',
 			},
-			url: '/static/admin',
+			url: getAppPath('/static/admin'),
 			conditions: [{ condition: 'user_is_admin' }],
 		},
 		webSections: [
@@ -98,7 +98,7 @@ export const connectAppDescriptor = {
 					value: 'Configure',
 				},
 				location: 'admin_plugins_menu/figma-addon-admin-section',
-				url: '/static/admin',
+				url: getAppPath('/static/admin'),
 				conditions: [
 					{
 						condition: 'user_is_admin',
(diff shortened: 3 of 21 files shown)


Roystbeef (approved):
Is it worth updating `env.example` and `env.test`? Maybe not since this is giving us assurance that if the environment variable wasn't set, everything's still working as expected?

knguyen-figma:
> Is it worth updating `env.example` and `env.test`? Maybe not since this is giving us assurance that if the environment variable wasn't set, everything's still working as expected?

I rather keep the existing env files and variables the same for now so that our deployed production service continues to work without any changes.

src/atlassian-connect.ts
```diff
@@ -110,29 +110,27 @@ export const connectAppDescriptor = {
...
 		jiraDesignInfoProvider: {
-			homeUrl: 'https://www.figma.com/',
+			homeUrl: getConfig().figma.webBaseUrl,
 			name: {
 				value: 'Figma',
 			},
 			key: 'figma-integration',
-			handledDomainName: 'figma.com',
-			logoUrl: `${getConfig().app.baseUrl}/static/figma-logo.svg`,
+			handledDomainName: getFigmaDomain(),
```
akostevich-atlassian, line 118: **issue:** The PR changes the value of `handledDomainName`. While originally it was set to `figma.com`, it is set to `www.figma.com` in this PR.

Unfortunately, it will be a breaking change since Atlassian platform currently handles `handledDomainName` as the primary domain. This platform behaviour seems non-ideal and, likely, needs to be fixed (**CC:** @jvrankul23). 

However, it will take some time, so I suggest reverting the value to `figma.com` for now.

admin/src/api/axiosInstance.ts
```diff
@@ -1,8 +1,10 @@
 import axios from 'axios';
+import { getAppBasePath } from '../utils';
 
 const THIRTY_SECONDS_IN_MS = 30_000;
 const axiosRest = axios.create({
 	timeout: THIRTY_SECONDS_IN_MS,
+	baseURL: getAppBasePath() ?? undefined,
```
akostevich-atlassian, line 7: **nitpick:** Since `getAppBasePath` returns a value of type `string | undefined`, we could remove ` ?? undefined` from here.

src/atlassian-connect.ts
```diff
@@ -110,29 +110,27 @@ export const connectAppDescriptor = {
...
+			logoUrl: getAppUrl('/static/figma-logo.svg'),
 			documentationUrl:
 				'https://help.figma.com/hc/en-us/articles/360039827834-Jira-and-Figma',
 			actions: {
 				getEntityByUrl: {
-					templateUrl: `${getConfig().app.baseUrl}/entities/getEntityByUrl`,
+					templateUrl: getAppUrl('/entities/getEntityByUrl'),
```
akostevich-atlassian, line 124: **question:** @knguyen-figma , do we need to change the base URL or base path on production?

I am asking since any configuration changes that cause changes in the Connect Descriptor requires a new release (in addition to deployment) to be applied. If this is the case, we need a multi-step migration to ensure that the app does not get in a broken state between a deployment and new release.

However, if there will no effective change on production (but only for Figma's dev and staging environments), no action is required here.

knguyen-figma: It's only for Figma's dev and staging environments!

admin/src/utils/parseTeamIdFromFigmaUrl.ts
```diff
@@ -1,5 +1,11 @@
 const VALID_FIGMA_ORIGINS = ['www.figma.com', 'figma.com'];
 
+if (import.meta.env.VITE_FIGMA_FOR_JIRA_FIGMA_WEB_DOMAIN) {
+	VALID_FIGMA_ORIGINS.push(
+		import.meta.env.VITE_FIGMA_FOR_JIRA_FIGMA_WEB_DOMAIN,
```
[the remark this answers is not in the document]
knguyen-figma: This is only used for internal development at Figma. Our development Figma environments have unique domain names per developer. I'll add a comment.

src/atlassian-connect.ts
knguyen-figma: I'll add a new env var to `config.ts` that falls back to `figma.com`:  `readEnvVarString('FIGMA_DOMAIN', 'figma.com')`

src/config/config.ts
```diff
@@ -40,6 +41,7 @@ export const getConfig = (): Config => {
 			app: {
 				baseUrl: readEnvVarString('APP_URL'),
 				key: readEnvVarString('APP_KEY'),
+				basePath: readEnvVarString('APP_BASE_PATH', ''),
```
akostevich-atlassian, line 44: **suggestion:** We could also consider simply using single `APP_URL` (which might contain a path as well) instead of `APP_URL` + `APP_BASE_PATH`.

Here is how we could achieve this:

1. On internal environments, set the `APP_URL` env variable to the required value (e.g., `https://figma-internal.com/dev`).
2. Make `config.app.baseUrl` of type `URL`.
  ```ts
  export const getConfig = (): Config => {
	  if (!config) {
		  config = {
			  app: {
				  baseUrl: new URL(readEnvVarString('APP_URL')),
				  key: readEnvVarString('APP_KEY'),
			  },
  			...
		  };
	  }
```
4. Use `getConfig().app.baseUrl.pathname` where the path is required. For example:
  ```ts
  // app.ts
  app.use(getConfig().app.basePath.pathname, rootRouter);
  ```
  
  ```ts
  // `getAppPath` is not required anymore as `getAppUrl("/test").pathname` can be used instead.
  export function getAppUrl(relativeUrl: string): URL {
	  return new URL(path, config.app.baseUrl);
  }
  ```

akostevich-atlassian (changes requested):
Please, see the comments.

src/config/config.ts
knguyen-figma: `baseUrl` in `atlassian-connect.json` did not play well when `getConfig().app.baseUrl` had a path. 

Also the URLs in `jiraDesignInfoProvider` are absolute but the other modules' URLs are relative so it was easier to just provide the extra `basePath` environment variable separately.

akostevich-atlassian: Great point, which raises an important question: are these "relative URLs" actually relative to the app domain or the path where the Connect Descriptor is available?

Have you tried, @knguyen-figma , to install the app from this branch in Jira with the `APP_BASE_PATH` path being used?

akostevich-atlassian: According to the [docs](https://developer.atlassian.com/cloud/jira/platform/connect-app-descriptor/#lifecycle), it is actually relative to URL specified in the `connectAppDescriptor.baseUrl`.

And from internal docs:
> Each add-on must have a unique `baseUrl`. If you would like to serve multiple add-ons from the same host, consider adding a path prefix into the `baseUrl`.

Therefore, the question here is "do we want `baseUrl` to include both `APP_URL` and `APP_BASE_PATH`? From my perspective, it is a good idea as it matches where the app is available. Considering this, I still think that using a single `APP_URL` might be beneficial.

PR #257, opened 2025-06-15
VULN-1636254 bump node version to 24.2.0

jvrankul23:
Bumping node to latest to resolve vulnerabilities.


akostevich-atlassian (changes requested):
Jon, please, also update the version in the following files:
- In `.github/workflows/actions.yml`. Ideally, we should use the version specified in the `.nvm` file instead of the hard-coded one.
- in `package.json`. There are 2 places in this file:
   - In `engines.node`
   - In `devDependencies["@types/node"]`

akostevich-atlassian (approved):

PR #258, opened 2025-07-01
NONE: add launchdarkly feature flagging support

Isaiah-Turner:
This PR adds a LaunchDarkly client so that we can feature flag new additions to the integration.

This PR adds the `CLUSTER_NAME` env variable so that we can use the appropriate secret in Figma's different environments. The use of `LAUNCH_DARKLY_SECRET_KEY` for production-like environments and `LAUNCH_DARKLY_SECRET_KEY_MAP` for local development matches how other Figma services connect to LaunchDarkly.

Testing:
- Set up devbox 
- Add some temporary logging in server.ts; make sure flag value correctly aligns with LaunchDarkly value
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
 				"core-js": "^3.36.0",
 				"dd-trace": "^4.17.0",
 				"express": "^4.18.2",
+				"launchdarkly-node-server-sdk": "^7.0.4",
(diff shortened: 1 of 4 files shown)


Roystbeef (approved):
o - we literally just add it. this is lit

PR #259, opened 2025-07-29
[NONE] create migration for figma file webhooks

Isaiah-Turner:


prisma/schema.prisma
```diff
@@ -71,3 +72,25 @@ model FigmaTeam {
   @@unique([teamId, connectInstallationId])
   @@map("jira_figma_team")
 }
+
+model FigmaFileWebhook {
```
akostevich-atlassian, line 76: **suggestion (non-blocking):** We could also consider storing webhook subscriptions in different tables (one per a webhook type), so they can be scaled and evolved separately. 

However, if we think that both `FILE_UPDATE` and `DEV_MODE_STATUS_UPDATE` belong to the same family of events (File-based events) and managed in the same way, the currently solution is perfectly fine!

akostevich-atlassian (approved):

prisma/schema.prisma
Isaiah-Turner: Good point! IMO these events are similar enough that we probably can treat and manage them in the same table

PR #260, opened 2025-07-30
[NONE]: add and delete file webhooks when designs are attached or removed

Isaiah-Turner:


package.json
```diff
@@ -33,6 +33,7 @@
 		"test:unit": "DOTENV_CONFIG_PATH=.env.test jest -c jest.config.unit.ts"
 	},
 	"dependencies": {
+		"@figma/rest-api-spec": "^0.33.0",
```
Isaiah-Turner, line 36: I replaced some of our custom types with types from our [official rest api spec](https://github.com/figma/rest-api-spec) where possible

src/domain/entities/figma-file-webhook.ts
```diff
@@ -0,0 +1,50 @@
+import type { ConnectUserInfo } from './connect-user-info';
+
+export enum FigmaFileWebhookEventType {
+	FILE_UPDATE = 'FILE_UPDATE',
+	DEV_MODE_STATUS_UPDATE = 'DEV_MODE_STATUS_UPDATE',
+}
+
+export class FigmaFileWebhook {
```
akostevich-atlassian, line 8: **suggestion (npn-blocking):** It seems we over-engineered a bit `FigmaFileWebhook`. This class is just holds data and does not have any business logic associated with it.

We could define `FigmaFileWebhook` as a read-only type without using classes.

```ts
export enum FigmaFileWebhookEventType {
	FILE_UPDATE = 'FILE_UPDATE',
	DEV_MODE_STATUS_UPDATE = 'DEV_MODE_STATUS_UPDATE',
}

export type FigmaFileWebhook = {
	readonly id: string;
	readonly webhookId: string;
	readonly webhookPasscode: string;
	readonly fileKey: string;
	readonly eventType: FigmaFileWebhookEventType;
	readonly createdBy: {
		readonly atlassianUserId: string;
		readonly connectInstallationId: string;
	};
};

export type FigmaFileWebhookCreateParams = {
	readonly webhookId: string;
	readonly webhookPasscode: string;
	readonly fileKey: string;
	readonly eventType: FigmaFileWebhookEventType;
	readonly createdBy: {
		readonly atlassianUserId: string;
		readonly connectInstallationId: string;
	};
};
```

src/infrastructure/figma/figma-client/types.ts
```diff
@@ -102,27 +102,6 @@ export type DeleteDevResourceRequest = {
 	readonly accessToken: string;
 };
 
-export type CreateWebhookRequest = {
```
akostevich-atlassian, line 105: **praise:** This is so good!

src/usecases/on-design-associated-with-issue-use-case.ts
```diff
@@ -54,5 +72,93 @@ export const onDesignAssociatedWithIssueUseCaseParams = {
...
+			connectInstallationId,
+		);
+	if (existingWebhooks.length > 0) {
+		return;
+	}
+
+	try {
+		const webhookPasscode = uuidv4();
+		const { fileWebhook, devModeStatusUpdateWebhook } =
+			await figmaService.createFileContextWebhooks(fileKey, webhookPasscode, {
```
akostevich-atlassian, line 118: **issue:** The current behaviour can cause creating "dangling" webhooks in an edge case when the first webhook is successfully created but the creation of the second one fails. This being the case, no webhook will be recorded in our database. As a result, when an operation is retried, the implementation will try to create a new set of webhooks.

In order to minimise the risk of this, please, consider the following

1. Split `FigmaService.createFileContextWebhooks` into 2: `FigmaService.createFileUpdateWebHook` and `FigmaService.createDevModeStatusUpdateWebHook`.
2. Update this code to do the following:
    1. Check whether the webhook of type `FILE_UPDATE` exists.
    2. If not, all `FigmaService.createFileContextWebhooks` and then `FigmaFileWebhookRepository.upsert` for the created webhook; otherwise -- skip.
    3. Check whether the webhook of type `DEV_MODE_STATUS_UPDATE` exists.
    4. If not, all `FigmaService.createDevModeStatusUpdateWebHook` and then `FigmaFileWebhookRepository.upsert` for the created webhook; otherwise -- skip.

akostevich-atlassian:
**praise:** This was a quite large and challenging work, @Isaiah-Turner ! Really love the changes and what they bring to the product. Great job!

PR #261, opened 2025-08-01
src: handle incoming file webhook events and dev mode status payloads

Isaiah-Turner:
This PR improves how we handling incoming webhooks events. To support file webhooks, we now search the `figmaFileWebhookRepository` in addition to the`figmaTeamRepository`.

To support `DEV_MODE_STATUS_UPDATE` webhooks, we added them as a case to our switch statement. Currently, we run the same `handleFigmaFileUpdateEvent` logic when we receive a dev mode status update.

### Testing
- This PR adds and updates existing tests
- To test manually, link a single section or frame in a file to an issue. Then, update the frame's dev mode status. In Jira we should see both badges appear.

https://github.com/user-attachments/assets/2e783c42-64f3-40c3-a268-b53a17a7ae51


src/web/routes/figma/figma-router.ts
akostevich-atlassian, line 1: **suggestion:** Similar to https://github.com/atlassian-labs/figma-for-jira/pull/260/files#r2266671913, please, consider updating integration tests for the `POST /figma/webhook` endpoint (or a new `POST /figma/webhook/file` if you decide to adopt the suggestion below).

Likely, they are currently passing since the Launch Darkly client falls back to `false` for the feature gate.

src/web/routes/figma/figma-router.ts
```diff
@@ -41,7 +46,43 @@ figmaRouter.post(
...
+				return res.sendStatus(HttpStatusCode.Ok);
+			}
+			default:
+				return res.sendStatus(HttpStatusCode.Ok);
+		}
+	},
+);
+
+// Handles FILE_UPDATE and DEV_MODE_STATUS_UPDATE webhook events for a specific file
+figmaRouter.post(
```
akostevich-atlassian, line 67: **praise:** Thank you for making this change! It looks really simple and clean!

akostevich-atlassian (approved):

PR #263, opened 2026-02-02
NONE: Ensure designs are synced only once on a webhook event

akostevich-atlassian:
### Changes

- **refactor:** Ensures that a design is fetched from _Figma_ and submitted to _Jira_ only **once** on a webhook event. Currently, a design that is linked to multiple _Work Items_ in _Jira_ is fetched and submitted multiple times.


src/common/array-utils.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Returns an array of unique items based on the given comparator.
+ */
+export function uniqueBy<T>(
```
akostevich-atlassian: **context:** Considering that `Set` cannot be used to de-deduplicate objects, add a utility method to handle this use case (e.g., to de-deduplicate a list of `FigmaDesignIdentifier`).

src/domain/entities/figma-design-identifier.ts
```diff
@@ -70,10 +70,7 @@ export class FigmaDesignIdentifier {
 		return [this.fileKey, this.nodeId].filter(Boolean).join('/');
 	}
 
-	equals(designIdentifier: FigmaDesignIdentifier): boolean {
-		return (
-			this.fileKey === designIdentifier.fileKey &&
-			this.nodeId === designIdentifier.nodeId
-		);
+	equal(other: FigmaDesignIdentifier): boolean {
```
akostevich-atlassian, line 73: **context:** This method was unused but it is needed now :) Slightly refactored it and covered with tests.

src/usecases/handle-figma-file-update-event-use-case.ts
```diff
@@ -95,8 +96,11 @@ async function syncDesignsToJira(
 
 	if (!associatedFigmaDesigns.length) return;
 
+	const atlassianDesignIds = associatedFigmaDesigns.map((x) => x.designId);
+	const uniqueDesignIds = uniqueBy(atlassianDesignIds, (x, y) => x.equal(y));
```
akostevich-atlassian, line 100: **context:** This is the core improvement, which prevents designs linked to multiple _Work Items_ in _Jira_ to be fetched from _Figma_ and submitted to _Jira_ multiple times. 

It should significantly reduce the amount of IO produced by the app on both _Figma_ and _Atlassian_ systems.

src/common/array-utils.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Returns an array of unique items based on the given comparator.
+ */
+export function uniqueWith<T>(
```
akostevich-atlassian, line 4: **context:** Considering that `Set` cannot be used to de-deduplicate objects, add a utility method to handle this use case (e.g., to de-deduplicate a list of `FigmaDesignIdentifier`).

Inspired by https://lodash.com/docs#uniqBy.

Isaiah-Turner (approved):

